PR #58, opened 2014-12-27
Added mouse support (mouse_show_cursor(), mouse_is_cursor_visible(), mouse_move() and sys_check_for_event())

bastienleonard:
I'm not sure what's the convention for porting names from the C API to Rust, so I pretty much kept them the same.

`sys_check_for_event()` is a bit messy internally, I think it should delegate much of its behavior to `KeyState` and `MouseState`.
I'm not too happy about the way it's used currently either. It doesn't feel like you're using a high level langage and requires to setup one or two mutable variables. I wasn't sure what to do so this is a direct port from the C API.
Here is an example of how I use it currently:

    let mut mouse_state = tcod::MouseState::new();
    let mut key_state = tcod::KeyState::new();

    loop {
        let event = tcod::sys_check_for_event(
            tcod::KEY | tcod::MOUSE,
            Option::Some(&mut key_state),
            Option::Some(&mut mouse_state));

        if !(event & tcod::KEY).is_empty() {
            println!("{}", key_state);
        } else if !(event & tcod::MOUSE).is_empty() {
            println!("{}", mouse_state);
        }

        if event == tcod::ANY {
            break;
        }
    }

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3,6 +3,7 @@
 extern crate libc;
 extern crate "tcod-sys" as ffi;
 
+extern crate core;
 use libc::{c_int, c_uint, c_float, uint8_t, c_void};
 
 pub use Console::Root as RootConsole;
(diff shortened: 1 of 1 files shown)


bastienleonard:
Seems like the history is cleaned up now? But I the only way I found was to rebase and then push with `--force`. Now there's a conflict, I may have missed something during the rebase, or I just don't understand something (I'm new to pull requests).

Either way, if you want me to look into the conflict, don't hesitate to give me write access.

tomassedovic:
Thanks! I think the recent change from `deriving` to `derive` was most likely causing the conflicts.

I've fixed it up locally. I also renamed `int` to `isize` to silence the warnings and I dropped the `Default` and `Show` traits for now.

The show implementation was not correct for flags. Unlike `enum`, `bitflags` can have any combination of the members. But the `Show` impl only worked for single values.

Anyway, I'm not against adding a proper Show for the flags (but it should then show somethingy like `KEY_PRESS | KEY_RELEASE`. I didn't want to block this PR on that. We can add that later.

Same for returning:

    enum Event {
        Key(KeyState),
        Mouse(MouseState)
    }

I prefer that much more to the Option<&...> interface, but it can be done in followup commits.

bastienleonard:
I'll submit improvements in a new pull request, I was wary of doing them before the merge.

What's the proper way of avoiding those unneeded pull request commits? Should I always rebase instead of merge on your repo?

tomassedovic:
Yeah, I understand.

Rebasing is the way to go. So if you want to start working on a pull request, you create a new branch from the current tcod-rs master. And if the master changes over time, you rebase to the current master before updating the pull request.

That does mean having to do a `--force` push, but that's fine since the branch will be deleted when the pull request is merged anyway.

On the other hand, if you wish to maintain your own fork and sync it with the upstream one periodically, you definitely want to do a merge in that situation. Because rebase messes with the history, so that'd make your fork unusable.

But again, if you then want to submit a pull request, it's always best to build it on the upstream master.

PR #72, opened 2015-01-09
Update to rust master

tomassedovic:
Works with `rustc 1.0.0-nightly (44a287e6e 2015-01-08 17:03:40 -0800)`

### DIFF
--- a/examples/astar-path-finding.rs
+++ b/examples/astar-path-finding.rs
@@ -5,7 +5,7 @@ extern crate tcod;
 use tcod::AStarPath;
 
 fn create_path() -> AStarPath<'static> {
-    let chess_board: [[int; 8]; 8] = [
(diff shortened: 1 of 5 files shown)


tomassedovic:
@GGalizzi do you want commit access, btw? Anyone who gets a patch merged is eligible if they want it :-).

GGalizzi:
Alrighty, yeah.

I'm very new to Rust btw, but for things like these with breaking changes I can help out when possible.

tomassedovic:
Yeah, that's completely fine :-). You should be set up now.

It's still good to use pull requests to have Travis check things and have someone else look at your code. Perfectly fine to ask others for suggestions if you're not sure about the particular approach, etc.

But yeah, this is super handy for things like fixing broken master or just adding missing libtcod functionality you need to use. I'm not always able to fix or merge things as soon as they break.

PR #74, opened 2015-01-10
Convert isize to i32

### LINKED ISSUE #73: Fix our use of integer types
We're currently using the old `int` / `isize` types for everything. This is causing a lot of casts throughout the codebase and is probably not the correct choice semanticaly.

I think that switching to `i32` makes the greatest amount of sense (and seems to be what most other gamedev libs are using).

tomassedovic:
Fixes #73

### DIFF
--- a/examples/astar-path-finding.rs
+++ b/examples/astar-path-finding.rs
@@ -1,11 +1,9 @@
-#![feature(unboxed_closures)]
-
 extern crate tcod;
 
 use tcod::AStarPath;
 
 fn create_path() -> AStarPath<'static> {
-    let chess_board: [[isize; 8]; 8] = [
+    let chess_board: [[i32; 8]; 8] = [
         [1, 0, 1, 0, 1, 0, 1, 0],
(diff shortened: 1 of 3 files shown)


tomassedovic:
Want to wait before merging this since I'm not entirely sure it's the right thing to do. I think it is, but there may be reasons why we want to keep using isize in our interfaces? Plus this changes the signature of almost all tcod functions.

@jgilchrist, @jaredonline, @pmsanford and anyone else: what do you think?

jgilchrist:
Looks sensible to me. The RFC regarding isize/usize [here](https://github.com/rust-lang/rfcs/blob/master/text/0544-rename-int-uint.md) implies that using isize/usize for anything which isn't pointer sized is bad practice, and i32 seems like a sensible default seeing as I believe it's the default fallback type for Rust integers now.

GGalizzi:
Agreed, using isize/usize for these is not the Rust way I believe. Using i32 seems like the way to go.

tomassedovic:
Thanks! I'll merge it now and release it as 0.4.0 on crates.io

PR #81, opened 2015-01-17
Add FOV functions.

GGalizzi:
Adds `Map::compute_fov` and `Map::is_in_fov` for calculating FOV.

I wasn't sure how to implement the different fov_algorithms, I hope this way is fine, I just did it as colors is done.

Please review~

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -6,6 +6,7 @@ use libc::{c_int, c_uint, c_float, uint8_t, c_void};
 
 pub use Console::Root as RootConsole;
 pub use ffi::TCOD_color_t as Color;
+pub use ffi::TCOD_fov_algorithm_t as FovAlgorithm;
 
 use std::num::FromPrimitive;
 use std::ffi::CString;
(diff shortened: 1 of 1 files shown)


tomassedovic:
Thanks! This is looking pretty good, but I think it'd be better to use `enum` instead of `bitflags!` here. Something like we do here:

https://github.com/tomassedovic/tcod-rs/blob/master/src/lib.rs#L671

What we do with colours is to export existing static structs because they have predefined values in libtcod and we want to carry those over. But when you're passing one value out of a specific set, enums make more sense, I think. Even the C version of libtcod has the algorithm defined as an enum:

http://roguecentral.org/doryen/data/libtcod/doc/1.5.1/html2/fov_compute.html?c=true&cpp=true

Do you mind rebasing and updating that? (I can do it myself, but don't want to mess with your PR).

GGalizzi:
Alright thanks.
I believe this should do it then :)

tomassedovic:
Sweet, thanks!

I like to have examples showing off new functionality so people can start with a working code and experiment with it. I'll whip one up at some point, but if you want to add it yourself, I'd definitely appreciate it :-)

PR #82, opened 2015-01-18
Added a simple mouse example

bastienleonard:


tomassedovic:
Nice, thanks! I've pushed this manually because I wanted to play with it and noticed a bit of a trailing whitespace.

PR #84, opened 2015-01-18
Update to work with the latest version of Rust

jgilchrist:
The bitflags! macro was moved out of Rust and onto crates.io, so this adds it as a dependency.


tomassedovic:
Thanks! I'll cut a release after I merge the other open PRs (in a few mins, hopefully).

PR #98, opened 2015-02-07
Test on both Linux and OSX

jgilchrist:
Yesterday I tried compiling my roguelike on OS X and got an error. This just makes sure that the library is tested on both platforms.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,9 @@
 language: rust
 
+os:
+  - linux
+  - osx
+
 before_install:
   - sudo apt-get update -qq
 


tomassedovic:
I've checked this site: http://docs.travis-ci.com/user/multi-os/ and it seems to require an email to enable multi-os testing. And the "Details" button still shows tcod being tested for linux.

I'll merge this and I've emailed the Travis folks. Hopefully we'll get the `os` feature enabled.

jgilchrist:
Looks like it's working, the test running for #99 is using Linux and OSX!

jgilchrist:
Looks like this could increase the time to run Travis significantly, the OSX build has been queued for over 15 minutes now :-1:

tomassedovic:
Thanks for the pull requests. I was sick and couldn't attend to tcod-rs. Which just shows that giving commit rights to contributors was the right thing to do :-).

tomassedovic:
By the way, where did you see the OSX tests being run?

Looking at this: https://travis-ci.org/tomassedovic/tcod-rs/builds/50559023

I only see the linux build. But yeah, it should have worked then, the Travis people enabled it.

jgilchrist:
Not a problem!

The OSX tests were being ran on Travis, but only for a couple of builds (e.g. [this one](https://travis-ci.org/tomassedovic/tcod-rs/builds/50551884) but I had to cancel it as the usage of `apt-get` was causing the builds to error.

If we want to have this tested on OSX we'll need to find an alternate way to get the dependencies.

tomassedovic:
Thanks, I can see the build now. For some reason (maybe because it was cancelled?) I don't see it in the list of builds. 

To make both OSes work, I think all we need to do is check `$TRAVIS_OS_NAME`. The OSX environment should have Homebrew installed, but I'm not sure if `language: rust` works there, too. If it doesn't, we can install it manually.

I'm not too happy about the OSX builds being queued up for a longer time, but a lot of tcod users have a mac so I wouldn't want to leave them behind. 15 minutes sounds unpleasant but not too bad.

jgilchrist:
You'll only see it in the list of pull request builds, because it never made it to the actual builds.

I believe `language: rust` works fine, so we'll just have to install the right dependencies using Homebrew. 15 minutes definitely isn't too bad.

Issue #102, opened 2015-02-18
Use a new Github API key for pushing docs


jgilchrist:
Currently, as I set up the automatic pushing of the documentation, Github sees me as having pushed to this repository whenever the documentation is updated (that is, whenever something is committed to master) - see [this](https://github.com/jgilchrist?tab=activity).

@tomassedovic is it possible we could switch to use an API key under your account? This will also prevent any breakage if for whatever reason I don't have contributor rights to this repository in future.

jgilchrist:
This doesn't appear to have worked. If you look at the expanded output at the bottom of the most recent build there's an error:

```
Collecting ghp-import
  Downloading ghp-import-0.4.1.tar.gz
Installing collected packages: ghp-import
  Running setup.py install for ghp-import
    changing mode of build/scripts-2.7/ghp-import from 644 to 755
    changing mode of /usr/local/bin/ghp-import to 755
Successfully installed ghp-import-0.4.1
remote: Anonymous access to tomassedovic/tcod-rs.git denied.
fatal: Authentication failed for 'https://@github.com/tomassedovic/tcod-rs.git/'
Done. Your build exited with 0.
```

tomassedovic:
Sigh. Maybe I gave it bad permissions. I'll try again a bit later.

tomassedovic:
Okay, it seems to be working now. I encrypted a wrong environment variable. It shows on my activity, the branch was updated and travis log no longer shows the error.

Closing again, feel free to reopen if it's still not all right.

jgilchrist:
Nope, it's all good. Thanks a lot!

PR #103, opened 2015-02-19
Use a new Github API key for the documentation

tomassedovic:
Closes: #102

(hopefully)


tomassedovic:
Hm, looks like Travis didn't download & run rustup.sh

I got this on master, too so it probably doesn't have anything to do with this. Let's wait a bit and see.

PR #111, opened 2015-02-24
Re-export special characters

jgilchrist:
I'm not sure that what I've done is the best way to achieve this.

I'm looking to be able to print special characters without adding `tcod-sys` as a dependency, so I've re-exported them in `tcod`. However, Rust doesn't let you cast to chars in constants, so I've had to cast them all to u8.

This means that to use them, there's an extra step:

```rust
console.print(loc, tcod::chars::VLINE as char);
// vs the more desirable:
console.print(loc, tcod::chars::VLINE);
```


jgilchrist:
I've also removed the TCOD_CHAR prefixes so that the paths don't duplicate any information:

```rust
tcod::chars::VLINE
// vs
tcod::chars::TCOD_CHAR_VLINE
```

Let me know your thoughts.

tomassedovic:
Good stuff. I'm happy to merge this as is. I have a couple of ideas on how to achieve the usability you mention but they both have some tradeoffs:

2. We could just export the constants as chars:

    pub const VLINE: char = '\u{C4}'

This means we have to duplicate the character codes, but I don't think that's necessarily a problem. The exported `Color` values do the same thing.

3. We could add a trait (`TcodPrintable` or whatever) that returns a `u32` and is implemented by `u8` and `char`. That would be easier to implement but it'd mean that our users would have to import that trait if they want to use `print` and friends.

Not sure there's another way to do this. Of the three options (assuming your's is number one) I'm leaning towards number two because it means we do a bit of extra work and our users won't have to care about anything. But I'm fine with any of these. What do you think?

jgilchrist:
I have to say, I didn't realise we'd be able to do your second option. Looks best to me, definitely the most ergonomic and it's a one time cost.

I'll see if I can get it done now and then we can merge it all as part of this PR.

tomassedovic:
All right, sounds good :-).

jgilchrist:
All done! There's a fair amount of noise going back and forth between types, so would you prefer me to squash the commits or would you like them all there for history's sake?

tomassedovic:
Nah, that's fine :-)

jgilchrist:
:fireworks:

PR #116, opened 2015-03-02
Correct tcod-sys version in its Cargo.lock

### LINKED ISSUE #115: tcod-sys version in Cargo.lock is only 2.0.0
For some reason, the `tcod-sys` `Cargo.lock` file is not updated if `tcod-rs` is compiled from the top directory. This means that the version listed for `tcod-sys` in its `Cargo.lock` is still 2.0.0 even though the crate is on version `2.0.5`.

jgilchrist:
Fixes #115

### DIFF
--- a/tcod-sys/Cargo.lock
+++ b/tcod-sys/Cargo.lock
@@ -1,4 +1,4 @@
 [root]
 name = "tcod-sys"
-version = "2.0.0"
+version = "2.0.5"
 


tomassedovic:
Cargo.lock shouldn't even be merged. We should remove & .gitignore it. Added an issue:

https://github.com/tomassedovic/tcod-rs/issues/119

PR #117, opened 2015-03-02
Remove env feature gate now it is stable

jgilchrist:


jgilchrist:
Merging now since this is a trivial change.

PR #120, opened 2015-03-02
Remove lock files

### LINKED ISSUE #119: Remove Cargo.lock from tcod-sys
Libraries should not check their Cargo.lock in:

http://doc.crates.io/guide.html#cargo.toml-vs-cargo.lock

So: we should delete Cargo.lock from tcod-sys and make sure it's in .gitignore.

jgilchrist:
Fixes #119


tomassedovic:
Thanks a lot!

PR #121, opened 2015-03-02
Move tcod-rs from std::{old_io, old_path} -> std::{io, path}

jgilchrist:
Seeing as Cargo and various other prominent projects have begun the migration onto the new IO and Path libraries, it seems like a good time for us to do the same.

I was as careful as I could be, but it'd be great if you could have a quick look over.

This is also a breaking change for our users - our public API takes old_path::Path objects so users will need to migrate onto the newer path API also.


tomassedovic:
Brilliant! It looks fine and appears to be correct in my testing. Merging :-)

PR #123, opened 2015-03-06
Added iterator version of tcod::system::check_for_event()

bastienleonard:
It's used like this:

```
for (flags, event) in tcod::system::check_for_event_iter() {
    match event {
        IteratedEvent::Key(ref key_state) => {
            // ...
        },
        IteratedEvent::Mouse(ref mouse_state) => {
            // ...
        }
    }
}
```

### DIFF
--- a/examples/mouse.rs
+++ b/examples/mouse.rs
@@ -11,20 +11,21 @@ fn main() {
     while !Console::window_closed() {
 
         loop {
-            let (_flags, event) = tcod::system::check_for_event(
-                tcod::KEY | tcod::MOUSE);
-
-            match event {
-                tcod::system::Event::Key(ref key_state) => {
-                    println!("{:?}", key_state);
-                },
-                tcod::system::Event::Mouse(ref mouse_state) => {
-                    x = mouse_state.cx as i32;
-                    y = mouse_state.cy as i32;
-                    println!("{:?}", mouse_state);
-                }
-                tcod::system::Event::None => {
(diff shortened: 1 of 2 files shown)


tomassedovic:
Looks great but I was wondering if we could name the iterator function to `tcod::system::events` and the event iterator struct to `Events`. That's how the standard library does it and it's much less typing :-).

I'd also like to merge `Event` and `IteratedEvent` to a single struct but I'm not sure how now, so let's not block it on that.

Are you okay with the renamings?

bastienleonard:
No problem with renaming.

I created `IteratedEvent` because `Event` contains a `None` member which makes no sense here.
We can probably merge them if we use `Option::None` instead of `Event::None`.

I'll submit both changes in this PR if that's OK with you.

tomassedovic:
Awesome, thanks!

bastienleonard:
In the end I made `sys_check_for_event` return an option. It doesn't seem that returning `(EventFlags, None)` would make any sense.

tomassedovic:
Yea, good point. I like the code. Can you fix the mouse example to fix the travis failure?

https://travis-ci.org/tomassedovic/tcod-rs/builds/53404667

I'd just fix it myself but I don't think I can add a commit to the pull request (the code is on your fork and I don't have an access to that obviously).

bastienleonard:
Should be good now!

tomassedovic:
\o/

PR #128, opened 2015-03-25
Rename `tcod-sys` to `tcod_sys`

### LINKED ISSUE #127: Hyphenation causes compilation failure on latest nightly
Since the [RFC regarding hyphens in crate names](https://github.com/lfairy/rfcs/blob/hyphens-considered-harmful/text/0000-hyphens-considered-harmful.md) has just merged, the hyphen in `tcod-sys` causes an error when attempting to build.

Cargo has been augmented to allow auto-translation between hyphens and underscores for package names, but we will have to rename the crate from `tcod-sys` to `tcod_sys` to allow it to compile properly.

jgilchrist:
In accordance with the new rules on naming crates.

This fixes the fact that currently, attempting to `cargo update` or `cargo build` on the latest rust fails.

It also lets us get rid of the quotes around:

```rust
// old
extern crate "tcod-sys" as ffi;

// new
extern crate tcod_sys as ffi;
```

@tomassedovic This is quite urgent as the library cannot be built on the latest Rust.

Fixes #127


tomassedovic:
Ooh, I didn't realise the change already landed. Thanks!

jgilchrist:
Not a problem! I think this should still work for publishing to crates.io seeing as we didn't rename the package, so it's probably worth releasing a new version.

tomassedovic:
On it.

PR #134, opened 2015-04-04
Manually fixed Copy and Clone generation

Gustorn:
rust-bindgen has not been updated yet, and the manual fix is fairly staightforward. In the two cases where adding the appropriate Clone attribute wasn't enough:
1. the generated LLVM IR seems to be fine for struct containing actual data (it calls memcpy on both arrays). (tcod_sys/lib.rs:486)
2. copying function pointers should be fine. (tcod_sys/lib.rs:626)

Also, all the provided examples work fine, but I don't have a bigger project to test it on right now.

### DIFF
--- a/examples/fov.rs
+++ b/examples/fov.rs
@@ -4,7 +4,7 @@ extern crate tcod;
 use tcod::{Console, BackgroundFlag, FovAlgorithm, Map};
 
 // We'll use a basic structure to define our tiles.
-#[derive(Copy)]
+#[derive(Copy, Clone)]
 pub struct Tile {
     ch: char,
     x: i32,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -799,7 +799,7 @@ impl Iterator for DijkstraPathIterator {
 
 
 #[repr(C)]
-#[derive(Copy)]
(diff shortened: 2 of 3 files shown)


tomassedovic:
Thanks! I *think* you can just put `*self` in the two manual implementations of `Clone`. I don't mind merging this since that code is an nigh-unreadable mess anyway, but if you want to make it shorter, I'll definitely be happier :-).

Lemme know what you think.

Gustorn:
I'll definitetely try it, I was doing it on autopilot... I'll see if it compiles.

tomassedovic:
\o/ Thanks!

@Gustorn do you want commit access? Anyone who gets a patch in is eligible :-).

Gustorn:
Sure, I'm planning on using the library for a hobby project, so I'll probably end up contributing whenever the latest rust breaks something (although that shouldn't happen that often now, I guess)

tomassedovic:
Cool, you should be all set now. Yeah, I hope the breakage is going to go down soon, but there's still a bunch of stuff missing and I don't want to block people on my limited availability (plus timezones are a thing).

PR #138, opened 2015-04-05
Module split to separate files

### LINKED ISSUE #136: Modules
I think it would be beneficial to split the project into modules for two reasons:

* if there'll be safe safe wrappers around everything in `libtcod` then a single `.rs` file will be really hard to handle
* it would probably make the life of new contributors  easier
* the module structure is already pretty well defined on the main project page, an example split:

| Name        |  Name       | Name    |
|----------------|---------------|------------|
| chars         | input          | system |
| colors        | map           |
| console     | pathfinding |

If we still want to have everything available from the top module we can just reexport the necessary `structs`, `traits` and `functions`.

Gustorn:
The only public API change is moving `Event` from `system` to `input`, since it seems to be completely input related. 

If something is really out of place I'll put it somewhere else.

Closes #136


tomassedovic:
Woo, nice! If you don't mind, I'll have a look tomorrow, My brain is about to shut down. But it's possible someone else will read through and merge it :-).

src/lib.rs
```diff
@@ -1,1717 +1,36 @@
-extern crate libc;
-extern crate time;
-extern crate tcod_sys as ffi;
+#![feature(path_ext)]
```
tomassedovic: This looks to have been missed by the merge with upstream. Opting into features is not supported in Rust's stable channel and I want to have tcod-rs run on 1.0.0. You should be able to remove this.

src/lib.rs
```diff
@@ -1,1717 +1,36 @@
...
+#![allow(dead_code)]
```
tomassedovic: I'd like this gone, too. It can hint on actual issues with the code (as is the case here, see show_cursor in input.rs).

src/input.rs
```diff
@@ -0,0 +1,292 @@
...
+    pub mbutton_pressed: bool,
+    pub wheel_up: bool,
+    pub wheel_down: bool,
+}
+
+pub mod mouse {
+    use bindings::ffi;
+    use bindings::c_bool;
+
+    fn show_cursor(visible: bool) {
```
tomassedovic: This should be `pub fn`. The reason the unused_code lint shows a warning here is because it's not exported and we're not using it anywhere.

src/system.rs
```diff
@@ -0,0 +1,108 @@
+extern crate std;
+extern crate time;
+
+use bindings::ffi;
+
+use self::time::Duration;
+use std::fs::PathExt;
```
tomassedovic: This should be removed once the PathExt feature is gone.

src/system.rs
```diff
@@ -0,0 +1,108 @@
...
+
+pub fn get_elapsed_time() -> Duration {
+    let ms: u32 = unsafe {
+        ffi::TCOD_sys_elapsed_milli()
+    };
+    return Duration::milliseconds(ms as i64)
+}
+
+pub fn save_screenshot(path: &std::path::Path) {
+    assert!(path.exists());
```
tomassedovic: This depends on the PathExt feature and this line should not actually be here. People should be able write to a nonexistent path, too. This was fixed in a previous commit but looks like the merge didn't work right.

tomassedovic:
Okay, this looks great! There's a couple of things I think need fixing before we merge this. I do this later today but if you just want to update this, I can just merge it. Or indeed, feel free to merge this yourself if you do end up fixing them :-).

I've a couple more ideas that can be added later but was wondering what you think:
1. the mouse mod in input.rs doesn't seem to make much sense to me now. Can we remove it?
2. Now that AStarPath and DijkstraPath are within the `pathfinding` namespace, I'm thinking of dropping the "Path" suffix since it's redundant.

Gustorn:
I think removing the `mouse` mod in input is a great idea, I just didn't want to do any sweeping changes during the initial split (this preserved backwards compatibility, except for the `system::Event -> input::Event` change). I'm also all for removing the `Path` suffix.

There's one thing though: both of these things would depend on, in my opinion, on not reexporting the contents of these modules: as it is now, `pathfinding` is not even a `pub mod`. If you'd like to have the reexports, then having it the way it is makes sense.

tomassedovic:
I'm okay with dropping the re-exports.

Gustorn:
Alright, I'll go ahead end do it. There is one workaround I ended up having to do: the `TcodMap` trait in the `map` mod: since `pathfinding` wants to use the native representation, and the member variable is not public (which is good). I think having it as a separate trait (and so having to explicitly include) is a good enough separation, but if you have any other ideas...

Gustorn:
Also, there's that comment in `pathfinding` about the `Drop` implementation. It lets me move `ffi::TCOD_path_t` into the the actual structs, implement `Drop`, and the compilation does not fail. Is there some kind of behaviour I'm still unaware of, that makes that approach impossible?

Gustorn:
Here are the changes both to the code and the examples. I left some reexports, because getting rid of them would've made the library pretty inconvenient to use (I can still get rid of them if you want).

tomassedovic:
Thanks! I like the `AsNative` trait.

PR #140, opened 2015-04-06
Console API rework

### LINKED ISSUE #137: Console API
While working on my project I noticed a few things about the `Console` API: the `Root` and `Offscreen` consoles seem to be clearly distinct concepts with some shared common functionality. The current API doesn't really reflect this: all functionality is implemented under the `Console` enum. Examples:

* `init_root` is only applicable to `Root` 
* `new` is only applicable to `Offscreen`
* all static methods could be either free functions or under `RootConsole`. Having them as free functions would probably provide a nicer API, while having them under RootConsole would match the `libtcod` documentation (which is why I'm more in favor of the 2nd option)

My suggestion would be the following: split Console into two structs and a trait:
* `OffscreenConsole`: struct for offscreen console functionality
* `RootConsole`: struct for the root console funcionality
* `Console`: trait for the shared functionality

In practice all trait methods except for `con(&self)` can be implemented in the `Console` trait itself, so `OffscreenConsole` and `RootConsole` would only have to provide `con(&self)`. 

Advantages:
* You can still use `&Console` or `Box<Console>` to have a collection of consoles, if you truly do not want to have distinctions between `Root` and `Offscreen` consoles.
* Type specific functionality will not bleed through to an other type (`new` and `init_root`).

Disadvantages:
* Breaking change
* More import statements (explicit `Console` trait import)

I'm only making these suggestions, because there's an open invitation on the main project page, if you want me to stop, please just say so.

Gustorn:
This still has the names `OffscreenConsole` and `RootConsole`. I'm kind of partial to these, since you can import them nicely from the `console` module and still know what they do.

If the naming is fine as it is, then this closes #137

### DIFF
--- a/examples/blit.rs
+++ b/examples/blit.rs
@@ -34,6 +34,6 @@ fn main() {
     console::blit(&trait_object, 0, 0, 20, 20, &mut root, 0, 20, 1.0, 1.0); 
     console::blit(&boxed_trait, 0, 0, 20, 20, &mut root, 20, 20, 1.0, 1.0); 
 
-    RootConsole::flush();
-    RootConsole::wait_for_keypress(true);
+    root.flush();
+    root.wait_for_keypress(true);
 }
--- a/examples/blit.rs
+++ b/examples/blit.rs
@@ -1,0 +1,39 @@
+extern crate tcod;
+
+use tcod::console;
+use tcod::{Console, RootConsole, OffscreenConsole};
+use tcod::colors;
+
+
+fn main() {
+    let mut root = RootConsole::init(80, 50, "Using blit with libtcod", false);
+    
+    let mut direct: OffscreenConsole = OffscreenConsole::new(20, 20);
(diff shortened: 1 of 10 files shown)


tomassedovic:
I'd prefer `Offscreen` and `Root` names in the `console` module. If someone wants the imported names to be more explicit, they can always say `use tcod::console::Root as RootConsole`.

And we can do that re-export in `lib.rs`, too. What do you think?

Otherwise, this looks great. For the record, making Console::con unsafe makes sense and I actually like that it's public -- people can use it to call the low-level FFI stuff if the want (which is inherently unsafe, too).

Gustorn:
Yes, I like the public `Console::con` as well, but it would've been more elegant if I could've implemented it using the `AsNative` traits. But it's not really feasible if we want to keep the current `blit` semantics (which are more important than the `unsafe Console::con`).

PR #142, opened 2015-04-08
Color naming

Gustorn:
Now they reflect the constant naming style suggested in the official Rust styleguide. If the use of snake case constants was deliberate then please just ignore and close this PR.


tomassedovic:
Yeah, good point. Thanks!

PR #143, opened 2015-04-08
Removed redundant native type wrappers.

Gustorn:
Rust allows the implementation of the Drop trait for structs with lifetime parameters, so this current solution is equivalent to the previous one.

### DIFF
--- a/src/pathfinding.rs
+++ b/src/pathfinding.rs
@@ -8,30 +8,20 @@ enum PathInnerData<'a> {
     Callback(Box<FnMut((i32, i32), (i32, i32)) -> f32+'a>, Box<(usize, usize)>),
 }
 
-// We need to wrap the pointer in a struct so that we can implement a
-// destructor. But we can't put ffi::TCOD_path_t directly inside AStar
-// because it includes the boxed closure which has a 'static lifetime so we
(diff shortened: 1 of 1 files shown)


tomassedovic:
I don't think we can remove the named lifetime in favour of `'static`. It works in the example, but not for move closures. In the astar example, I just moved the body of `create_path` into main and removed `move` from the closure (so it references its environment) and it no longer compiles because the closure isn't static anymore. And we definitely want to support non-static closures.

That is orthogonal to dropping the native wrapper. It wasn't possible to set a constructor on a generic struct (and lifetimes are generics) but that was fixed now so we can do that.

I.e. I'm all for merging the first commit, but the second one would be a regression. I'll add a path-finding example with a non-static closure when I have some time to make sure we don't regress.

Gustorn:
Yeah, sorry about that, I'll just revert the second commit. Also what do you think about having some of the examples as `#[test]` too?

tomassedovic:
Thanks! I thought about adding `#[test]` tests, but there isn't a lot to that wouldn't just test the underlying libtcod code. Things like our path walk iterators, the root console initialiser, etc. but not much more imho. But yeah, let's test those.

PR #144, opened 2015-04-09
Added Root console initializer

### LINKED ISSUE #137: Console API

### LINKED ISSUE #146: Setting custom fonts [regression]
As of #140 it is impossible to set a custom font for the console, since `set_custom_font` was converted to a method, but has to be called before we could possibly have an instance of the `Root` console.

Gustorn:
This implementation matches the latest version discussed in #137 with two modifications: 
1. newtype introduced for font paths, which makes all font-related arguments in the raw `Root` initializer explicit
2. added `Root::initializer()` static function to `Root`. This is to avoid an unnecessary import for `RootInitializer`. The downside is that `init` and `initializer` are fairly similarly named. 

Since this is a one-time function call, I don't think the added verbosity is too bad, and it's much nicer to work with.

Closes #137
Closes #146

### DIFF
--- a/examples/blit.rs
+++ b/examples/blit.rs
@@ -6,7 +6,10 @@ use tcod::colors;
 
 
 fn main() {
-    let mut root = RootConsole::init(80, 50, "Using blit with libtcod", false);
+    let mut root = RootConsole::initializer()
+        .size(80, 50)
+        .title("Using blit with libtcod")
+        .init();
     
     let mut direct: OffscreenConsole = OffscreenConsole::new(20, 20);
     let mut boxed_direct: Box<OffscreenConsole> = Box::new(OffscreenConsole::new(20, 20));
--- a/examples/chars.rs
+++ b/examples/chars.rs
@@ -4,7 +4,11 @@ use tcod::{Console, RootConsole};
 use tcod::chars;
 
 fn main() {
-    let mut root = RootConsole::init(80, 50, "Example of libtcod's special chars", false);
+    let mut root = RootConsole::initializer()
+        .size(80, 50)
+        .title("Example of libtcod's special chars")
+        .init();
+
     root.clear();
 
     // The top half of the box
--- a/examples/colors.rs
+++ b/examples/colors.rs
@@ -5,7 +5,11 @@ use tcod::colors;
 
 
 fn main() {
-    let mut con = RootConsole::init(80, 50, "Using colours with libtcod", false);
+    let mut con = RootConsole::initializer()
+        .size(80, 50)
+        .title("Using colours with libtcod")
+        .init();
+
     con.set_default_background(colors::darkest_green);
     con.set_default_foreground(colors::lighter_azure);
 
--- a/examples/fov.rs
+++ b/examples/fov.rs
@@ -13,9 +13,9 @@ pub struct Tile {
 }
 
 fn main() {
-    let mut root = RootConsole::init(40,40, "FOV example", false);
-    let mut map = Map::new(40,40);
+    let mut root = RootConsole::initializer().size(40, 40) .title("FOV example").init();
 
+    let mut map = Map::new(40,40);
     let mut tiles = Vec::new();
 
     root.clear();
--- a/examples/keyboard.rs
+++ b/examples/keyboard.rs
@@ -5,7 +5,11 @@ use tcod::input::Key::Special;
 use tcod::input::KeyCode::{Up, Down, Left, Right, Escape};
 
 fn main() {
-    let mut con = RootConsole::init(80, 50, "libtcod Rust tutorial", false);
+    let mut con = RootConsole::initializer()
+        .size(80, 50)
+        .title("libtcod Rust tutorial")
+        .init();
+
     let mut x = 40;
     let mut y = 25;
     while !con.window_closed() {
--- a/examples/minimal.rs
+++ b/examples/minimal.rs
@@ -3,7 +3,8 @@ extern crate tcod;
 use tcod::RootConsole;
 
 fn main() {
-    let mut root = RootConsole::init(80, 50, "Minimal libtcod loop", false);
+    let mut root = RootConsole::initializer().size(80, 50).title("Minimal libtcod loop").init();
+
     while !root.window_closed() {
         root.flush();
         let key = root.wait_for_keypress(true);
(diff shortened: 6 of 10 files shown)


Gustorn:
While I was looking through the `libtcod` docs and making the initializer, I noticed something strange about `console::FontFlags`/ I don't even know why it's handled as bitflags in the original lib, but here's my point of view: the `FONT_LAYOUT_*` flags are clearly exclusive, they describe a completely different set of layout rules for the font in question. This would be a nice candidate to move into an enum. As for `FONT_TYPE_GRAYSCALE` it basicly seems like an on-off switch, so we could just have a bool flag for it.

tomassedovic:
Cool (and sorry this took so long)! A suggestion: maybe we could use `builder` or `build` instead of `initializer`. It's what this pattern is called in general so that may make it clearer and fix the init/initializer similarity. But I don't mind `initializer` either.

Second,  I'm a bit conflicted about `FontPath`. With this patch, setting fonts is really inconsistent:
* `RootInitializer::font` takes `&str`
* `Root::set_custom_font` takes `&std::path::Path`
* `Root::init` takes `FontPath`.

Also, not sure how can one trigger #146 since you're required to set a font when you get a root console -- either through the initializer or init. We could theoretically just remove `Root::set_custom_font` now.

Regardless though, I think we should unify the way font paths are passed in and I think we should use `std::path::Path`. So either use FontPath(Path) or just take a naked Path in.

Since all the examples now use the builder pattern, I prefer to just pass &Path in.

What do you think?

tomassedovic:
Oh, and about FontFlags: what you're saying makes sense. I'm all for it. I just copied the way libtcod did it initially.

Gustorn:
I'm conflicted on the `initializer` -> `builder` rename, only because for me `builder` implies that you can make multiple copies, and this is not the case here. If you think it's for the best, I'll go and do the rename (for the record I'm not really liking the `initializer` name either: the `initializer`...`init` seems a bit awkward to me).

I'm all for getting rid of `FontPath` if we "advertise" `initializer` as the main method of creating a `Root` console (or we could even make `init` unsafe or private). If it remains in the public API then it's a coin toss for me: on one side there's the inconsistency you mentioned, on the other, removing `FontPath` would make that parameter the only "naked" one in `Root::init`. Even if that's the case, I'm not that woried about getting rid of it.

#146 references the current status quo: as of now (without this PR) `set_custom_font` takes a `&mut self`, but we can only get an instance of the `Root` console after initialization, after which calling it will not work. Having an initializer and/or taking the font parameters in `init` (so basicly this PR) solves the issue. 
Also about removing `set_custom_font`: I already made it private, because it makes the `Root::init` easier to read. I can copy-paste it inside init if you'd like, but I think it's fine as it is.

I'll go ahead and implement the `FontFlags` change and wait for the verdict on the other topics.

Gustorn:
Since the `FontFlags` split requires taking an additional parameter (`font_grayscale` is basicly a bool now), I'm much more in favor of getting rid of `Root::init`. A constructor taking 8 parameters is not a pleasant sight... If we get rid of it, we could just move all the initialization logic into `RootInitializer::init`, with a `builder` or `initializer` static function on `Root`.

tomassedovic:
Sounds good to me. Good point on builder implying multiple copies. So yeah,
let's keep it as is.

Gustorn:
There, it's all the change we discussed, with the following exception: I opted to declare a `FontType` enum for the `Grayscale` flag, in case `libtcod` decides to add more font types in the future. Other than that, this should be good now. I tested all the examples and they still all work (unfortunate, that the colors change blocks the auto-merge and the Travis build, but oh well).

jgilchrist:
It would be good to get this merged - unless I'm mistaken, it's currently impossible to properly set a custom font on the root console. That is, it has to look for `terminal.png` when it is created, even if it will never use it.

Gustorn:
Yeah that's what issue #146 is for. If you need to set a custom font, you can do it with the version that came before the Console API change: call `Console::set_custom_font` before `Console::init`.

jgilchrist:
Yeah, that's what I was doing up until the API change. I'll stick with `terminal.png` until this gets merged.

Gustorn:
You could fork and remove the `&mut self` from the `set_custom_font` if it proves the be a problem, but I think this should get merged pretty soon.

Gustorn:
@jgilchrist  I've thought up a workaround for you, if the font changing behaviour is somewhat urgent for you: just create an empty `Root` console, and since `Console::con` always returns `0` for the `Root` console you can just call `set_custom_font` on that. It's not a pretty solution, but it works.

And this actually highlights a weakness with using an empty struct for the `Root` console: it's creation is not limited. Is there a way to forbid the creation of objects that doesn't involve adding a private member variable?

tomassedovic:
Merged. We can improve this in future commits. Can't find anything better than a private field to prevent instantiating Root directly now.

I thought phantom types could do it, but I wasn't able to come up with an example that worked now. That may be worth looking into though.

PR #147, opened 2015-04-14
Root creation blocking and consistency

Gustorn:
The blocking of the Root initializer with phantom types (mentioned in #144) was indeed the right idea. The way it is implemeted now generates no extra assembly (which is to be expected from `PhantomData`) but still blocks explicit instantiation of `Root` consoles.

The second part of this PR is making the API of the `Color` struct consistent with that of `Map's`. Since the `AsNative` trait was already approved, I implemented that on `Color` instead of a method directly on he struct. The `FromNative` trait is a natural addition that matches the style of `AsNative` quite nicely. 

This has the following advantages:
* no more potentially unsafe operations exposed as safe (in `Color`'s case `TCOD_color_t` was safe anyway, but it's better to be safe than sorry).
* consisteny between modules: if someone wants to use the underlying native type, they'll have to import `AsNative` or `FromNative`
* no change in any other place in the public API

### DIFF
--- a/src/bindings.rs
+++ b/src/bindings.rs
@@ -16,6 +16,12 @@ pub trait AsNative {
     unsafe fn as_native(&self) -> <Self as AsNative>::Output;
 }
 
+pub trait FromNative {
+    type Input;
+
+    unsafe fn from_native(input: <Self as FromNative>::Input) -> Self;
+}
+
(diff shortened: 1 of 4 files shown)


tomassedovic:
Nice!

Issue #80, opened 2015-01-16
cargo test gives linker errors


jgilchrist:
For example:
```
---- engine::actor::Actor_0 stdout ----
	error: linking with `cc` failed: exit code: 1
note: cc '-Wl,--as-needed' '-m64' '-L' '/usr/local/lib/rustlib/x86_64-unknown-linux-gnu/lib' '-o' '/tmp/rustdoctest.r1BGY18VAAb0/rust-out' '/tmp/rustdoctest.r1BGY18VAAb0/rust_out.o' '-Wl,--whole-archive' '-lmorestack' '-Wl,--no-whole-archive' '-Wl,--gc-sections' '-pie' '-nodefaultlibs' '/home/jonny/code/rslike/target/librslike-33665a348a9c0d5a.rlib' '/home/jonny/code/rslike/target/deps/libtcod-835af025147faddf.rlib' '/home/jonny/code/rslike/target/deps/libtcod-sys-8caba187ba4adc8b.rlib' '-L' '/usr/local/lib/rustlib/x86_64-unknown-linux-gnu/lib' '-lstd-4e7c5e5c' '-L' '/home/jonny/code/rslike/target' '-L' '/home/jonny/code/rslike/target/deps' '-L' '/usr/local/lib/rustlib/x86_64-unknown-linux-gnu/lib' '-L' '/home/jonny/code/rslike/.rust/lib/x86_64-unknown-linux-gnu' '-L' '/home/jonny/code/rslike/lib/x86_64-unknown-linux-gnu' '-Wl,--whole-archive' '-Wl,-Bstatic' '-Wl,--no-whole-archive' '-Wl,-Bdynamic' '-ltcod' '-ltcod' '-ltcod' '-ldl' '-lpthread' '-lrt' '-lgcc_s' '-lpthread' '-lc' '-lm' '-lcompiler-rt'
note: /usr/bin/ld: cannot find -ltcod
/usr/bin/ld: cannot find -ltcod
/usr/bin/ld: cannot find -ltcod
collect2: error: ld returned 1 exit status

error: aborting due to previous error
thread 'engine::actor::Actor_0' panicked at 'Box<Any>', /home/rustbuild/src/rust-buildbot/slave/nightly-dist-rustc-linux/build/src/libsyntax/diagnostic.rs:144
```

For some reason, it's passing the flag `-ltcod` to the linker three times. This doesn't happen when building, only when testing.

tomassedovic:
[...]

PR #150, opened 2015-04-16
Documentation

Gustorn:
Some work on the `console` module documentation. I tried to proofread it a few times, but might've missed a few mistakes here and there. Also, I don't know if you want to have excessive documentation in the main developement branch. If not, then just create a new branch and I'll resubmit the PR. The docs could use some more examples for the individual methods (printing, etc.), but this is all I had time for now.

Also I tried to mirror `libtcod`'s style of "comment even the most obviously named function", even though I kind of disagree with the concept: `set_fullscreen` is pretty self-explanatory for example. So if you'd prefer it that way, I'll happily remove the redundant comments.

EDIT: So, #80 is still not fixed in cargo upstream, so we cannot do doctests if we want to have a documentation with examples. After the `More example fixes` commit, the only errors I was getting were the linker ones, so the examples should not have any bugs in them.

### DIFF
--- a/src/console.rs
+++ b/src/console.rs
@@ -1,9 +1,9 @@
 //! The console emulator handles the rendering of the game screen and the keyboard input
 //!
-//! It provides the necessary traits and types for working with the different console types in
-//! `tcod-rs`, which include the [Console](./trait.Console.html) trait and the
+//! It provides the necessary traits and types for working with the different console types, 
+//! including the [Console](./trait.Console.html) trait and the
 //! [Root](./struct.Root.html) and [Offscreen](./struct.Offscreen) console types.
-//! It's worth mentioning, that only one `Root` console may exist at any given time, and it has to
+//! It's worth mentioning that only one `Root` console may exist at any given time, and it has to
 //! be initialized at the start of the program.
 //!
 //! # Examples
@@ -67,12 +67,12 @@ use input::{Key, KeyPressFlags, KeyState};
 /// A type representing secondary consoles
 ///
 /// `Offscreen` consoles allow you draw on secondary consoles as you would on `Root` consoles, and then
-/// `blit` their contents onto other consoles (including `Root`). There are some limitations when
+/// `blit` their contents onto other consoles (including `Root`). There are some limitations
 /// compared to `Root` consoles, however:
 ///
 /// * Functions manipulating the main window or handling user input are limited to the `Root`
 /// console
-/// * `Offscreen` consoles may not be `flush`ed to the screen directly
+/// * `Offscreen` consoles may not be `flushed` to the screen directly
 ///
 /// # Examples
 ///
@@ -174,7 +174,7 @@ impl Offscreen {
 ///
 ///     root.clear();
 ///     // Output style manipulation
-///     // Calling the output functions on root
+///     // Calling the output functions
 ///     root.flush();
 /// }
 /// ```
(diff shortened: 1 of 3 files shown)


jgilchrist:
#80 was just fixed, we're good for doctests now!

jgilchrist:
(That said, the fix won't appear on the nightly until tomorrow).

Gustorn:
Also, I think the goal is to have this build on the beta too, which means no doctests until the next major release.

jgilchrist:
Ah, never mind in that case :+1:

src/console.rs
```diff
@@ -264,49 +525,54 @@ pub trait Console {
...
     }
 
+    /// Sets the console's default foreground color. This is used in several printing functions. 
     fn set_default_foreground(&mut self, color: Color) {
         unsafe {
             ffi::TCOD_console_set_default_foreground(self.con(), color.as_native());
         }
     }
 
-    fn console_set_key_color(&mut self, color: Color) {
```
tomassedovic: lol nice catch

Gustorn: Oh yeah, that probably shouldn't have gone into the docs PR, I just happened to notice it...

tomassedovic: No worries. I'm reading through and will merge manually once I'm done. Barring any actual problems.

Gustorn:
Here's a git patch to fix the examples for the tuple-based `blit` (not that it's terribly hard to do by hand).

```patch
From 07bc6af1fcda15322159cd54271a7924db45a25a Mon Sep 17 00:00:00 2001
From: Gustorn <gustorn@gmail.com>
Date: Thu, 16 Apr 2015 23:07:48 +0200
Subject: [PATCH 8/8] Updated docs for tuple-based blit

---
 src/console.rs | 10 +++++-----
 1 file changed, 5 insertions(+), 5 deletions(-)

diff --git a/src/console.rs b/src/console.rs
index 4785865..37bd67b 100644
--- a/src/console.rs
+++ b/src/console.rs
@@ -96,7 +96,7 @@ use input::{Key, KeyPressFlags, KeyState};
 ///     let mut root = Root::initializer().init();
 ///
 ///     let mut direct = Offscreen::new(20, 20);
-///     console::blit(&direct, 0, 0, 20, 20, &mut root, 0, 0, 1.0, 1.0);
+///     console::blit(&direct, (0, 0), (20, 20), &mut root, (0, 0), 1.0, 1.0);
 /// }
 ///
 /// ```
@@ -771,10 +771,10 @@ pub trait Console {
 ///     let mut trait_object: &Console = &Offscreen::new(20, 20);
 ///     let mut boxed_trait: Box<Console> = Box::new(Offscreen::new(20, 20));
 ///
-///     console::blit(&direct, 0, 0, 20, 20, &mut root, 0, 0, 1.0, 1.0);
-///     console::blit(&boxed_direct, 0, 0, 20, 20, &mut root, 20, 0, 1.0, 1.0);
-///     console::blit(&trait_object, 0, 0, 20, 20, &mut root, 0, 20, 1.0, 1.0);
-///     console::blit(&boxed_trait, 0, 0, 20, 20, &mut root, 20, 20, 1.0, 1.0);
+///     console::blit(&direct, (0, 0), (20, 20), &mut root, (0, 0), 1.0, 1.0);
+///     console::blit(&boxed_direct, (0, 0), (20, 20), &mut root, (20, 0), 1.0, 1.0);
+///     console::blit(&trait_object, (0, 0), (20, 20), &mut root, (0, 20), 1.0, 1.0);
+///     console::blit(&boxed_trait, (0, 0), (20, 20), &mut root, (20, 20), 1.0, 1.0);
 /// }
 ///
 /// ```
-- 
2.3.5

```

tomassedovic:
Okay, this is awesome. Thanks!

I haven't seen any obvious mistakes and we can fix anything that comes up.

For the record -- documentation is great and should be on master. I just didn't get to it (and lately, I've been having very little time so again thanks for doing all this). It's a bit weird here since libtcod has its own documentation, but 1) the UI is much worse than what rust doc produces and 2) we did change some stuff up and not everyone is familiar with libtcod anyway.

Merging.

jgilchrist:
There's an incremental update to the beta (1.0.0-beta.2) which includes the fix for doctests. If this is the new compatibility target then we can re-enable our doctests.

Gustorn:
It is working, but I'm not sure if it's a good idea to enable them: a few examples (ones showcasing input handling for example) actually pop up a console, waiting for a keypress to exit. This is some pretty annoying behaviour to deal with in my opinion. I'd suggest running the doctests locally, but having them disabled in `master`.

jgilchrist:
I ran them locally, and I see what you mean. One of them also ran into a stack overflow and they don't seem to finish executing.

If we really want them enabled, one option would be to ignore the problematic tests.

```
/// ```ignore
/// Test goes here
/// ```
```

That said, all of the tests so far seem to be testing the console, so it seems like it might be best to leave them disabled for now.

Gustorn:
Which one ran into the stack overflow? I don't think that's supposed to happen. I can see the input handling one waiting for input, and there's the basic structure one, that's basicly an infinite loop at this point. But SO?

jgilchrist:
I'm not sure which test is causing the crash, but this is the crash information:

```
Crashed Thread:        1  Dispatch queue: com.apple.libdispatch-manager

Exception Type:        EXC_CRASH (SIGSEGV)
Exception Codes:       0x0000000000000000, 0x0000000000000000

Thread 0:: Dispatch queue: com.apple.main-thread
0   libstd-4e7c5e5c.dylib         	0x0000000107f7e5c5 sys::stack_overflow::imp::signal_handler::h536527a47190450ahAv + 757
1   libsystem_platform.dylib      	0x00007fff904dcf1a _sigtramp + 26
2   ???                           	000000000000000000 0 + 0
3   libtcod.dylib                 	0x00000001083d7ea1 TCOD_sys_init + 913
4   libtcod.dylib                 	0x00000001083b3109 TCOD_console_init + 265
5   rust_out                      	0x0000000107f218e7 console::RootInitializer$LT$$u27$a$GT$::init::h877f08955986d7bfCPa + 695
6   rust_out                      	0x0000000107f1dd92 main::had319cd9bf9e742efaa + 146
7   libstd-4e7c5e5c.dylib         	0x00000001080099b9 rust_try_inner + 9
8   libstd-4e7c5e5c.dylib         	0x00000001080099a6 rust_try + 6
9   libstd-4e7c5e5c.dylib         	0x0000000107f808f9 rt::lang_start::h6b294fc2544f55d4UKw + 697
10  rust_out                      	0x0000000107f1ddef main + 79
11  libdyld.dylib                 	0x00007fff922a85c9 start + 1
```

It also doesn't happen every time I run the tests.

When it crashes, it leaves a process called `rust_out` which has a very high CPU utilisation.

Gustorn:
Going by the disassembly it is this one:

```rust
use tcod::console::Root;

fn main() {
    let mut root = Root::initializer().init(); // Replace with custom initialization code

    while !root.window_closed() {
        // Handling user input
        // Updating the gamestate
        // Rendering the results
    }
}
```

The interesting part is: this doesn't overflow the stack on my machine (Linux), but it also does not acknowledge `CTRL+C` as an interrupt signal. Well, not much I can do about it, this example is meant to be an infinite loop.

PR #152, opened 2015-04-17
Started implementing missing features

Gustorn:
I'm not too attached to the macros part, I pretty much just wanted to get a second opinion, and I thought I would put it in the PR. I think it's nicer to use then the miriad of printing functions in `libtcod` and it also has the advantage of accepting Rust-style format strings. 

The thing I dislike about it is that the printing modifiers come before the string we want to print. Because of the macro definition rules there's only one solution to this problem: wrapping the `$($arg: tt)*` part in some kind of parentheses, which is also not ideal.

I'll do [these](http://doryen.eptalys.net/data/libtcod/doc/1.5.2/html2/console_advanced.html) next, any ideas for a nice API? I'm thinking enums with variants that take `i32`'s as their parameters and a single `Console::draw` method. The problem with this is that not all functions have the `clear` flag. Any way to easily get around this? Just add a bool member to the enumu variants?

### DIFF
--- a/examples/print_macros.rs
+++ b/examples/print_macros.rs
@@ -1,0 +1,28 @@
+#[macro_use]
+extern crate tcod;
+
+use tcod::{Console, RootConsole, BackgroundFlag, TextAlignment};
+
+fn main() {
+    let mut root = RootConsole::initializer().size(80, 50).title("Displaying text").init();
+
+    // No optional parameters
+    tcod_print!(root, At(1, 1), "Any text params may be {}", "formatted with println! formatting");
+
+    // One optional parameter
+    tcod_print!(root, At(1, 3), Wrap(6, 2), "Simple wrap"); 
+    tcod_print!(root, At(1, 6), Bg(BackgroundFlag::None), "No background flag"); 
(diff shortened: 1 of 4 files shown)


tomassedovic:
That's a big macro you've got there :-).

I'm kind of okay with the order being inconsistent. It's a macro so I favour ease of typing over consistency with the other printing functions. +1 from me.

Not sure I understand what you're proposing with the advanced printing functions. Can you give an actual example? Personally, I'd just go with the separate functions but I'm sure there's a nicer way to do this.

tomassedovic:
I've restarted the build, feel free to merge if it passes.

Gustorn:
I'd revised my opinions on the advanced printing functions, since we seem to (somewhat) mirror the C API, and they are inconsistent enough that they need their separate functions. The original proposal would've been something like this:
```rust
enum Shape {
    Rect(i32, i32),
    VLine(i32),
    HLine(i32),
   ...
}

fn draw(&mut self, shape: Shape, flag: BackgroundFlag) { ... }
```

I think it would still be nicer, but unfortunately the `clear` argument that some of the functions take would make this awkward to implement and use.

PR #162, opened 2015-06-05
Unicode - Ascii printing dispatch

Gustorn:
Please don't merge this yet, I'm sending this mainly because I'd like some feedback on the API and whatnot. So as it stands, this succesfully dispatches `&str` and `String` to the Unicode, and `&[u8]` to the ASCII printing functions. 

Problems
-------------
1. The type-by-type implementation is tedious, but I haven't found a way to make a nice blanket implementation. Since there are no negative trait bounds I don't see a simple way around this problem. If anyone has any ideas, feel free to post them.
2. Is there a more efficient way to convert a `&str` to a `*const i32`? The current method works, but allocating a new `Vec` for each string seems like a bit of an overkill. (Also do we have to explicitly push a `\0` at the end of these `Vecs`?).
3. I can't figure out a way to simply accept a byte literal in the print functions: I have to add an `as &[u8]` for it to work. 
4. There are two possible API designs:

  * **Take the text we want to print by value:** this makes it nicer to directly print `&str` values (no need for an explicit `&`), but will move `String` values. Note that this is equivalent to the current implementation, since we didn't support passing `Strings` earlier. The only problem is that you cannot simple do an `as_ref` on the `String`, since it has both `AsRef<str>` and `AsRef<[u8]>` implemented, leading to an ambigous call... On this note, I really wish Rust would differentiate between the `&str` and `&[u8]` representations. 
  * **Take it by reference:** this makes it more annoying to print `&str` (and so string literals), since you have to explicitly add a `&` to every one of them, but has the advantage of not moving `String` values.

Alternatives:
-----------------
1. We could just drop the whole UTF thing, assert on [AsciiExt::is_ascii()](https://doc.rust-lang.org/nightly/std/ascii/trait.AsciiExt.html#method.is_ascii) and keep doing things as they are now. If I understand correctly this is pretty much what the Python bindings do (the `libtcod` docs say that UTF is not supported with them).

2. We could dispatch at runtime based on `is_ascii`, so we only do the `&[u8]` to UTF conversion when it's absolutely needed. This is probably the most flexible solution, but loses out in raw performance with ASCII strings. The thing is that the implementation in this PR heavily favors the use of UTF strings, which will have worse performance (due to the string-by-string `Vec` allocations).

3. Have `libtcod` style `_utf` functions. I don't really like this option. It doesn't make for a very user-friendly API: we have these explicit functions when we could've dispatched them either at runtime or compile time. The only advantage is being able to use `&str` in both functions (not possible with this PR, possible with `Alt #2`).

#### Opinion
After giving it some thought, I actually think that `Alternative #2` is probably the best option here. It makes it easy to use the interface correctly. The printing of ASCII strings is probably the most common use-case, so making it easy should be a top priority. With the runtime dispatch we add a single `if` overhead, but retain correctness (in the UTF case we would've had to do the conversions anyway). This PR was actually pretty fun to experiment with (and provided some great experience for my hobby project!), but it's being too tricky for a library API in my opinion. If there's a decision I'd be happy to implement whatever it ends up being.

cc @P1start since they seemed to be interested in the whole conversion.

### DIFF
--- a/examples/blit.rs
+++ b/examples/blit.rs
@@ -16,6 +16,7 @@ fn main() {
     let mut trait_object: &Console = &OffscreenConsole::new(20, 20);
     let mut boxed_trait: Box<Console> = Box::new(OffscreenConsole::new(20, 20));
 
+
     root.set_default_background(colors::DARKEST_GREEN);
 
     direct.set_default_background(colors::RED);
--- a/examples/blit.rs
+++ b/examples/blit.rs
@@ -16,6 +16,7 @@ fn main() {
     let mut trait_object: &Console = &OffscreenConsole::new(20, 20);
     let mut boxed_trait: Box<Console> = Box::new(OffscreenConsole::new(20, 20));
 
+    trait_object.obj_unsafe(&9);
 
     root.set_default_background(colors::DARKEST_GREEN);
 
--- a/examples/blit.rs
+++ b/examples/blit.rs
@@ -16,8 +16,6 @@ fn main() {
     let mut trait_object: &Console = &OffscreenConsole::new(20, 20);
     let mut boxed_trait: Box<Console> = Box::new(OffscreenConsole::new(20, 20));
 
-    trait_object.obj_unsafe(&9);
-
     root.set_default_background(colors::DARKEST_GREEN);
 
     direct.set_default_background(colors::RED);
--- a/examples/roguelike-tutorial-2.rs
+++ b/examples/roguelike-tutorial-2.rs
@@ -51,12 +51,12 @@ impl Object {
         }
     }
 
-    pub fn draw(&self, con: &mut Console) {
+    pub fn draw(&self, mut con: &Console) {
         con.set_default_foreground(self.color);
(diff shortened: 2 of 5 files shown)


tomassedovic:
Thanks for looking into this alternative, too.

I think I'm leaning towards the second alternative as well. If the runtime cost gets too high, we can re-evaluate. Plus the users can always use the FFI functions (though I don't want to have that as an answer for every difficult problem :-) ).

Just checking: the reason why he have to do the Vec -> *i32 dance is because Rust's strings are UTF-8 but libtcod's aren't?

If we do go with the alternative 2. we should be able to keep the same interface, right?

Gustorn:
The `Vec` -> `i32` dance is necessary, because Rust stores Unicode characters as an array of `[u8]s`, but `libtcod` expects `wchar_t` (which is just wrong... it should be `uint32_t`). 

And yes, we can keep the same interface with Alt #2.

tomassedovic:
Cool. Let's go with alternative #2 then. You wanna look into it? I can do it but it'll be a few days before I have the time.

Gustorn:
Closing in favor of #163

PR #163, opened 2015-06-11
Unicode printing and AsRef consistency

Gustorn:
Implementation as per #161 and #162

As far as I could tell everything works fine. The example mentioned in #161 (printing `b"90\xf8"`) also works correctly, provided we use the Unicode equivalent of the special character: `"90\u{f8}"`. It also seems like there is no need to implement separate printing for single characters: `TCOD_console_put_char_xx` and `TCOD_console_set_char` already take `i32` as their arguments, which `char` can be easily casted to (as it is done now).

### DIFF
--- a/src/console.rs
+++ b/src/console.rs
@@ -56,6 +56,7 @@
 
 extern crate std;
 
+use std::ascii::AsciiExt;
 use std::marker::PhantomData;
 use std::path::Path;
 
@@ -335,9 +336,9 @@ impl Root {
     }
 
     /// Sets the main window's title to the string specified in the argument.
-    pub fn set_window_title(&mut self, title: &str) {
+    pub fn set_window_title<T>(&mut self, title: T) where T: AsRef<str> {
         unsafe {
-            let c_title = CString::new(title.as_bytes()).unwrap();
+            let c_title = CString::new(title.as_ref().as_bytes()).unwrap();
             ffi::TCOD_console_set_window_title(c_title.as_ptr());
         }
     }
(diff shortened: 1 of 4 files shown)


src/console.rs
```diff
@@ -815,12 +852,12 @@ pub trait Console {
...
     /// using inverted colours.
-    fn print_frame(&mut self, x: i32, y: i32, width: i32, height: i32,
-                   clear: bool, background_flag: BackgroundFlag, title: Option<&str>) {
+    fn print_frame<T>(&mut self, x: i32, y: i32, width: i32, height: i32,
+                     clear: bool, background_flag: BackgroundFlag, title: Option<T>) where Self: Sized, T: AsRef<str> {
         assert!(x >= 0 && y >= 0 && width >= 0 && height >= 0);
         assert!(x + width < self.width() && y + height < self.height());
         let c_title: *const c_char = match title {
-            Some(s) => CString::new(s).unwrap().as_ptr(),
+            Some(s) => CString::new(s.as_ref()).unwrap().as_ptr(),
```
tomassedovic: It's not explicitly said in the docs, but I assume the title is ASCII as well. Maybe we want to assert is_ascii here?

src/console.rs
```diff
@@ -711,60 +712,96 @@ pub trait Console {
...
+        if text.is_ascii() {
             let c_text = CString::new(text.as_bytes()).unwrap();
-            ffi::TCOD_console_print_rect(self.con(), x, y, width, height, c_text.as_ptr());
+            unsafe {
+                ffi::TCOD_console_print_rect(self.con(), x, y, width, height, c_text.as_ptr());
+            }
+        } else {
+            let c_text = text.chars().collect::<Vec<_>>();
+            unsafe {
+                ffi::TCOD_console_print_utf(self.con(), x, y, c_text.as_ptr() as *const i32);
```
tomassedovic: Shouldn't this be `TCOD_console_print_rect_utf`?

tomassedovic:
Thanks! Spotted a typo that needs fixing and an assert we may want to add.

Fell free to merge this once you fix the print_rect_utf stuff (or I'll do it when I get to it).

tomassedovic:
Sweet, thanks!

PR #164, opened 2015-06-13
Support for proper printing of ASCII literals

Gustorn:
I wrote a detailed summary of my opinions in #161, but here's a possible implementation for the ASCII literal printing. This actually works quite well, and as soon as Rust has negative trait bounds we can implement `TcodString` for `AsRef<str>` as well. The obvious limitation is the hard coded limit on the fixed size arrays, but note that even the standard library only implements traits for fixed-sized arrays up to  the [size 32](https://doc.rust-lang.org/nightly/core/primitive.array.html).

Other than that I think this works as most users would expect: ASCII literals in Rust guarantee that for each character a single byte is passed (with a value of 0-255) which is what `libtcod` expects. In the case of regular strings the UTF conversion occurs if the given string had a character that was outside of the original (0-127) ASCII range. 

Closes #161

### DIFF
--- a/examples/text.rs
+++ b/examples/text.rs
@@ -6,9 +6,9 @@ fn main() {
     let mut root = RootConsole::initializer().size(80, 50).title("Displaying text").init();
 
     root.print_ex(1, 1, BackgroundFlag::None, TextAlignment::Left,
-                  "Text aligned to left.");
+                  b"Ascii text aligned to left.\xf8");
     root.print_ex(78, 1, BackgroundFlag::None, TextAlignment::Right,
-                  "Text aligned to right.");
+                  "Unicode text aligned to right.\u{f8}");
     root.print_ex(40, 15, BackgroundFlag::None, TextAlignment::Center,
                   "And this bit of text is centered.");
     root.print_ex(40, 19, BackgroundFlag::None, TextAlignment::Center,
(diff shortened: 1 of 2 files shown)


tomassedovic:
Thanks!

This does work as I'd expect (wish we could erase the size of the byte arrays instead of having 30 impls, but it's not the end of the world).

I've noticed that printing a `String` (without dropping it afterwards) involves an explicit deref:

    root.print_ex(40, 22, BackgroundFlag::None, TextAlignment::Center, &*format!("Borrowed string."));

whereas previously, the asterisk wasn't necessary:

    root.print_ex(40, 22, BackgroundFlag::None, TextAlignment::Center, &format!("Borrowed string."));

I assume that's because of the lack of negative bounds? Not a big deal and I'm happy to merge this as-is but I wonder why that is and my brain is a bit sluggish now.

Gustorn:
Yeah, that was because `AsRef<str>` is blanket implemented for `&T`, and I forgot to implement `TcodString` for `&String`. With negative trait bounds this will be automatically fixed, with:

```rust
impl<T> TcodString for T where T: AsRef<str> + !AsciiLiteral {
    fn as_ascii(&self) -> Option<&[u8]> {
        match self.is_ascii() {
            true => Some(self.as_ref()),
            false => None,
        }
    }
}
```

tomassedovic:
Agreed.

Issue #165, opened 2015-06-14
tcod-sys doesn't build on OSX


ProfessionalNihilist:
Hey, having some trouble building tcod on OSX.

```
anathema@Adorjan ~/C/r/ferl> cargo build
   Compiling tcod-sys v2.0.11
failed to run custom build command for `tcod-sys v2.0.11`
Process didn't exit successfully: `/Users/anathema/Code/rust/ferl/target/debug/build/tcod-sys-2a137dae9d1f93a4/build-script-build` (exit code: 101)
--- stdout
[...]

PR #166, opened 2015-06-15
Add `Console::get_height_rect`

tomassedovic:
This can only be used with ASCII strings, unfortunately.

### DIFF
--- a/examples/text.rs
+++ b/examples/text.rs
@@ -13,6 +13,10 @@ fn main() {
                   "And this bit of text is centered.");
     root.print_ex(40, 19, BackgroundFlag::None, TextAlignment::Center,
                   "Press any key to quit.");
+
+    let wrapped_text = "This text is wrapped to form X lines: https://xkcd.com/688/";
+    let lines = root.get_height_rect(10, 25, 22, 10, wrapped_text);
+    root.print_rect(10, 25, 22, 10, wrapped_text.replace("X", &lines.to_string()));
     root.flush();
     root.wait_for_keypress(true);
 }
--- a/src/console.rs
+++ b/src/console.rs
@@ -898,6 +898,27 @@ pub trait Console : AsNative<ffi::TCOD_console_t> {
         }
     }
 
+    /// Compute the height of a wrapped text printed using `print_rect` or `print_rect_ex`.
+    ///
+    /// While the print functions accept strings with non-ASCII characters,
+    /// there is no unicode equivalent for this function in libtcod. You must
+    /// pass a text with ASCII characters only.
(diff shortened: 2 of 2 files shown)


Gustorn:
We could technically modify this to work with UTF strings: if the string passed in is not ASCII then we build a new, ASCII string with the same length as `text.chars()`, so something like this:

```rust
match text.as_ascii() {
    Some(text) => { ... },
    None => {
        let text = str::from_utf8(text.as_ref()).unwrap().chars().map(|_| "X").collect::<String>();
        // Pass this to the FFI
    }
}

tomassedovic:
That's a pretty cool idea (though `TCOD_console_get_height_rect` looks at whitespace so we'd have to be a bit smarter than that). But I'm also a bloody idiot and libtcod already has `TCOD_console_get_height_rect_utf` which I've never noticed even when I was specifically looking for it...

Gustorn:
Well, it's pretty much just a copy-paste from the other ASCII-UTF functions then.

tomassedovic:
Yep :-). I expected it to be there but I still didn't see it before.

Gustorn:
I asked this in the other thread, but do you think we should push a `\0` at the end of the UTF converted `Vec`? I haven't looked too much into it, but it might only work because Rust zeroes the memory after the characters? Or are Rust strings zero terminated as well?

tomassedovic:
Yeah, I forgot to answer that.

Rust's strings are not zero-terminated and can actually contain `\0` inside them. Converting them to `CString` turns them into proper zero-terminated C strings.

But yeah, we should push a zero at the end of the vector we're creating. Unless there is a `\0` in the string explicitly, `chars` won't add it. And yeah this is a memory safety issue that we were just lucky not to hit yet I think.

Gustorn:
Yeah, this is my bad, and it IS pretty bad. The memory that Vec allocated is zeroed, but if we have the same number of characters as it allocated then the C function will happily write till the next 0. A nice [example](http://is.gd/8ooqE9) for the bug.

tomassedovic:
Yeah :/. I wanted to cut a new release later today to fix the issues with #165 plus we've added a bunch of new stuff, but we need to fix this first.

Got a time now? I'll do it in ~6 hours otherwise.

Gustorn:
Yeah, sent the PR. Also before posting a new release could you take a look at #165? The inclusion of `libz.a` in the crates.io crate is incredibly strange, and I think that's the reason the build fails on OS X.

tomassedovic:
Thanks. And yeah I definitely plan to look into that first.

PR #167, opened 2015-06-15
Fixed memory safety bug

Gustorn:


src/console.rs
```diff
@@ -579,6 +579,13 @@ impl<T> TcodString for T where T: AsRef<[u8]> + AsciiLiteral {
     }
 }
 
+#[inline]
+fn to_wstring(text: &[u8]) -> Vec<char> {
+    let mut ret = str::from_utf8(text.as_ref()).unwrap().chars().collect::<Vec<_>>(); (text.as_ref());
```
tomassedovic: What's the `(text.as_ref())` at the end for? Looks like a no-op to me.

Gustorn: And that's what I get for copy-pasting the original line.

tomassedovic:
Looks like travis has trouble reaching crates.io. Restarted the build.

PR #170, opened 2015-06-16
Image toolkit

Gustorn:
Safe wrapper for `libtcod's` [image toolkit](http://roguecentral.org/doryen/data/libtcod/doc/1.5.2/html2/image.html). I left a few TODOs in the code, if you think they can be ignored I'll resubmit the PR without them, but I think they're worth considering. I'll comment on the appropriate lines with a little bit of further information.

### DIFF
--- a/src/image.rs
+++ b/src/image.rs
@@ -189,35 +189,30 @@ impl Image {
     }
 }
 
-//TODO: Unify blit in the wrapper?
-// console::blit: (source, source_coords, dest, dest_coords)
-// image::blit_rect: (source, dest, source_coords, dest_dimensions, flag)
-// image_blit: (source, dest, source_coords, flag, source_scale, source_angle)
-//
-// Worst offender compared to console::blit
-// image::blit_2x(source, dest, dest_coords, source_coords, dest_dimensions)
-
-pub fn blit_rect<T>(src: &Image, dest: &T, (x, y): (i32, i32), (width, height): (i32, i32), flag: BackgroundFlag) where T: Console {
-    assert!(x >= 0 && y >= 0 && x < src.width && y < src.height);
-    assert!(width >= -1 && height >= -1);
+pub fn blit_rect<T>(src: &Image, (width, height): (i32, i32),
+                    dst: &mut T, (x, y): (i32, i32), flag: BackgroundFlag) where T: Console {
+    assert!(width >= -1 && height >= -1 && width <= src.width && height <= src.height);
+    assert!(x >= 0 && y >= 0 && x < dst.width() && y < dst.height());
     unsafe {
-        ffi::TCOD_image_blit_rect(src.tcod_image, *dest.as_native(), x, y, width, height, flag as u32);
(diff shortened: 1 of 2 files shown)


src/image.rs
```diff
@@ -0,0 +1,223 @@
...
+    assert!(width >= -1 && height >= -1);
+    unsafe {
+        ffi::TCOD_image_blit_rect(src.tcod_image, *dest.as_native(), x, y, width, height, flag as u32);
+    }
+}
+
+pub fn blit<T>(src: &Image, dest: &T, (x, y): (i32, i32), flag: BackgroundFlag, (scale_x, scale_y): (f32, f32), angle: f32) where T: Console {
+    assert!(x >= 0 && y >= 0 && x < src.width && y < src.height);
+    unsafe {
+        //TODO: libtcod docs say that (x, y) are ints, but the FFI expects floats. Mirror the docs
```
Gustorn: This one is pretty straightforward: there is a discrepancy between the public `libtcod` docs and their actual API. Which one should we follow?

src/image.rs
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    pub fn invert(&mut self) {
+        unsafe {
+            ffi::TCOD_image_invert(self.tcod_image);
+        }
+    }
+}
+
+//TODO: Unify blit in the wrapper?
```
Gustorn: These blits pretty much scream for a nicer API now. I have no really good ideas, since macros don't know the types of their arguments... But even if there are a lot of differnet `blit` functions, this API design on `libtcod's` part is ludicrous. I really don't mind any `dest, src` ordering, as long as it is consistent. This is anything but.

In lieu of a nice `blit` API alternative, I'm in favor of some kind of consistent argument order in the `blit` functions.

tomassedovic: Yeah :/.

Definitely agree on making the order consistent. Maybe we can figure out a nicer blit API later, can't think of anything now.

src/image.rs
tomassedovic: _"This function allows you to specify the **floating point coordinates** of the center of the image, its scale and its rotation angle."_

http://roguecentral.org/doryen/data/libtcod/doc/1.5.2/html2/image_blit.html?c=true#1

And the signature in the code is floats, too:

https://github.com/tomassedovic/tcod-rs/blob/master/tcod_sys/libtcod/src/image_c.c#L274
https://github.com/tomassedovic/tcod-rs/blob/master/tcod_sys/libtcod/src/image.cpp#L76

I'd go with floats

Gustorn: I'll change it to floats, I was looking at the C signature in the docs :/

src/image.rs
Gustorn: My suggestion: `source, source_args, dest, dest_args`. This is consistent with the current Rust style (src, dest), and it helps to distinguish the heap of float and int parameters at the end of the image blit functions. It is also what `console::blit` uses. 

As for an other API, I have no idea. I think syntax extensions could make the type-specific dispatch happen, but I don't have the time to dig into that right now.

Gustorn:
It's incredible how much a little organization can reveal how nonsensical my original asserts were.

PR #171, opened 2015-06-17
Name Generator

Gustorn:
Name generation functionality, but I had to wrap to whole pseudorandom-generator module as well (namegen, noise requires this). The name generator has a very obvious limitation: you can only ever have one name generator at any given time. I have no idea how to wrap a global construct more elegantly, but we do need a separate structure to remain memory safe: `libtcod` would happily use random number generators that have already been freed (hence the stored copies in `Namegen`).

If there are any alternative API ideas, I'd be happy to hear them.

### DIFF
--- a/examples/namegen.rs
+++ b/examples/namegen.rs
@@ -1,10 +1,16 @@
  extern crate tcod;
 
 use tcod::namegen::Namegen;
-use tcod::random::Rng;
+use tcod::random::{Rng, Algo};
+
+fn setup_namegen() -> Namegen {
+    let rng = Rng::new(Algo::MT);
+    let mut namegen = Namegen::new().unwrap();
+    namegen.parse_with_rng("examples/names.txt", &rng);
+    namegen
+}
 
 fn main() {
-    Namegen::parse("examples/names.txt", Rng::get_instance());
-    println!("{}", Namegen::generate("king").unwrap());
-    Namegen::reset();
+    let namegen = setup_namegen();
+    println!("{}", namegen.generate("king").unwrap());
 }
--- a/examples/namegen.rs
+++ b/examples/namegen.rs
@@ -1,0 +1,10 @@
+ extern crate tcod;
+
+use tcod::namegen::Namegen;
+use tcod::random::Rng;
+
+fn main() {
+    Namegen::parse("examples/names.txt", Rng::get_instance());
+    println!("{}", Namegen::generate("king").unwrap());
+    Namegen::reset();
+}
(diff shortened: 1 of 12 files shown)


examples/namegen.rs
```diff
@@ -0,0 +1,16 @@
+ extern crate tcod;
+
+use tcod::namegen::Namegen;
+use tcod::random::{Rng, Algo};
+
+fn setup_namegen() -> Namegen {
+    let rng = Rng::new(Algo::MT);
+    let mut namegen = Namegen::new().unwrap();
+    namegen.parse_with_rng("examples/names.txt", &rng);
```
tomassedovic: `examples/names.txt` is not in the pull request and running the example without it seems to panic. An omission?

src/namegen.rs
```diff
@@ -0,0 +1,101 @@
...
+use bindings::{AsNative, c_char};
+use random::Rng;
+
+static mut NAMEGEN_FREE: bool = true;
+lazy_static! {
+    static ref NAMEGEN_MUTEX: Mutex<()> = Mutex::new(());
+}
+
+pub struct Namegen {
+    rng: Vec<Rng>,
```
tomassedovic: Having a vector here is because you can call `TCOD_namegen_parse` multiple times with different random generators and they all need cleanup?

examples/namegen.rs
Gustorn: Yes, I was on autopilot and did `git add examples`, but it was in `.gitignore`. It's fixed now.

tomassedovic:
Looks good. I can't think of a better API now -- this seems sane enough. Not happy about the mutex but I see the point.

Eventually someone will write a better library that doesn't have to deal with libtcod's baggage and is much more Rusty, but until then, we'll make do :-).

src/namegen.rs
Gustorn: `TCOD_namegen_generate` will actually try and use the PRNG that was passed in and with Rust those will be automatically dropped on scope exit, so for example the example I posted simply segfaulted with a memory corruption error. It's fairly easy to reproduce: just use `*rng.as_native()` in one of the `parse` functions and the example should crash.

I actually did it an other way as well: you could just limit the `Rng`s passed in based on their lifetimes. It's much less elegant and pretty annoying to use, hence the current solution.

Gustorn: The lifetime limited version would look something like this:

```rust
pub struct Namegen<'a> {
    _lifetime: &'a PhantomData<()>
}

impl<'a> Namegen<'a> {
    pub fn parse_with_rng<T>(&mut self, path: T, rng: &'a Rng) where T: AsRef<Path> {
        let path_string = CString::new(path.as_ref().to_str().unwrap()).unwrap();
        unsafe {
            ffi::TCOD_namegen_parse(path_string.as_ptr(), *rng.as_native());
        }
    }
}
```

PR #182, opened 2015-08-22
Two overly cautious asserts

tomob:
In `print_frame` the assertion is off by one.
In `blit` negative destination is perfectly fine.

### DIFF
--- a/src/console.rs
+++ b/src/console.rs
@@ -942,7 +942,7 @@ pub trait Console : AsNative<ffi::TCOD_console_t> {
     fn print_frame<T>(&mut self, x: i32, y: i32, width: i32, height: i32,
                      clear: bool, background_flag: BackgroundFlag, title: Option<T>) where Self: Sized, T: AsRef<str> {
         assert!(x >= 0 && y >= 0 && width >= 0 && height >= 0);
-        assert!(x + width < self.width() && y + height < self.height());
+        assert!(x + width <= self.width() && y + height <= self.height());
         let c_title: *const c_char = match title {
             Some(s) => {
                 assert!(s.as_ref().is_ascii());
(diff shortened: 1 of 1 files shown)


Gustorn:
I'm somewhat vary of relying `libtcod` for refusing negative values in blit. It happens to work now, but it's not bulletproof. From the `libtcod` code:

```c
if ( (unsigned) cx >= (unsigned) src->w || (unsigned)cy >= (unsigned) src->h ) continue;
if ( (unsigned) dx >= (unsigned) dst->w || (unsigned)dy >= (unsigned) dst->h ) continue;
```

These are the two lines that make negative destinations work. They only do because 2's complement happens to work out during the conversion. With either a very small negative number or a very high console width these can be easily bypassed. I also don't see the value of doing this: whenever you'd want a negative destination you could just specify a smaller source, and it would do the same thing (with less work).

As for `print_frame`: that assert is in accordance with the [docs](http://doryen.eptalys.net/data/libtcod/doc/1.5.2/html2/console_advanced.html?c=false&cpp=false&cs=false&py=false&lua=false#3). Looking at the `C` source you're actually right: `x + width` is never used (only `x + width - 1` or `- 2`). The question is: go with the docs or the source?

tomob:
Well, I'm porting samples_cpp.cpp to Rust and both modification are needed to make it work. 

    static TCODConsole secondary(SAMPLE_SCREEN_WIDTH/2,SAMPLE_SCREEN_HEIGHT/2); // second screen
    // ...
    if (! init ) {
        init=true;
        secondary.printFrame(0,0,SAMPLE_SCREEN_WIDTH/2,SAMPLE_SCREEN_HEIGHT/2,false,TCOD_BKGND_SET,"Offscreen console");
        // ...

The frame fills the console exactly. 

Blit:

    x+=xdir;y+=ydir;
    if ( x == SAMPLE_SCREEN_WIDTH/2+5 ) xdir=-1;
    else if ( x == -5 ) xdir=1;
    if ( y == SAMPLE_SCREEN_HEIGHT/2+5 ) ydir=-1;
    else if ( y == -5 ) ydir=1;
    // ...
    // blit the overlapping screen
    TCODConsole::blit(&secondary,0,0,SAMPLE_SCREEN_WIDTH/2,SAMPLE_SCREEN_HEIGHT/2,
				&sampleConsole,x,y,1.0f,0.75f);

Both `x` and `y` intentionally go below zero.

Gustorn:
Alright, I'm fine with the change then, I'll just make a PR for the original `libtcod` project to make nevative blits actually safe and to fix the docs.

Gustorn:
Nevermind, blit is actuallt safe, the checks are just scattered all over the place.

tomob:
Thank you.

PR #185, opened 2015-08-27
Implementing samples

tomob:
Implemented 7 (out of 11) examples from libtcod's samples_cpp.cpp. It required a small change to console.rs (adding two wrappers for C API).


tomassedovic:
Thanks! I'll have a look later today.

tomob:
No rush.

Gustorn:
This is really well done, I only have some small nitpicks (I'll comment inline).

examples/samples.rs
```diff
@@ -0,0 +1,1175 @@
...
+    dirr : [i8; 4],
+    dirg : [i8; 4],
+    dirb : [i8; 4],
+    rng : Box<ThreadRng>,
+}
+
+impl ColorsSample {
+    fn new() -> Self {
+        ColorsSample {
+            cols : [Color {r:50,  g:40, b:150},
```
Gustorn: `Color::new()` would be nicer here I think. If you think it's not clear enough, we could just rename it to `Color::from_rgb()`, to match `Color::from_hsv()`

examples/samples.rs
```diff
@@ -0,0 +1,1175 @@
...
+                    Color {r:240, g:85, b:5},
+                    Color {r:50,  g:35, b:240},
+                    Color {r:10,  g:200, b:130}],
+            dirr : [1, -1, 1, 1],
+            dirg : [1, -1, -1, 1],
+            dirb : [1, 1, 1, -1],
+            rng: Box::new(rand::thread_rng()),
```
Gustorn: I don't think you have to `Box` the rng.

examples/samples.rs
```diff
@@ -0,0 +1,1175 @@
...
+              first: bool,
+              event: Option<(EventFlags, Event)>) -> ();
+}
+
+struct ColorsSample {
+    cols : [Color; 4],
+    dirr : [i8; 4],
+    dirg : [i8; 4],
+    dirb : [i8; 4],
+    rng : Box<ThreadRng>,
```
Gustorn: I don't think you have to `Box` the rng.

examples/samples.rs
```diff
@@ -0,0 +1,1175 @@
...
+            BottomRight,
+        };
+
+        // ==== scan the whole screen, interpolating corner colors ====
+	    for x in 0..SAMPLE_SCREEN_WIDTH {
+		    let xcoef = (x as f32) / ((SAMPLE_SCREEN_WIDTH-1) as f32);
+
+		    // get the current column top and bottom colors
+		    let top = colors::lerp(self.cols[Dir::TopLeft as usize],
+                                   self.cols[Dir::TopRight as usize],
```
Gustorn: I am slightly OCD for this kind of stuff, but could you fix the formatting here (and in the other `lerp`)?

examples/samples.rs
```diff
@@ -0,0 +1,1175 @@
...
+        let dx = 0.0;
+        let dy = 0.0;
+        let di = 0.0;
+        if self.torch {
+            // TODO implemenent when noise in wrapped in Rust API
+        }
+
+        self.display_map(console, dx, dy, di);
+
+        match event {
```
Gustorn: An `if let` would remove one level of nesting here.

examples/samples.rs
```diff
@@ -0,0 +1,1175 @@
...
+
+    fn recalculate(&mut self) -> () {
+        if self.using_astar {
+            self.astar.find((self.px, self.py), (self.dx, self.dy));
+        } else {
+            self.dijkstra_dist = 0.0;
+            self.dijkstra.compute_grid((self.px, self.py));
+            iterate_map(&mut |x, y, _c| {
+                let d = self.dijkstra.distance_from_root((x, y));
+                if d.is_some() && d.unwrap() > self.dijkstra_dist {
```
Gustorn: A `match` would be much more idiomatic here I think:

```rust
match d {
    Some(d) if d > self.dijkstra_dist => self.dijkstra_dist = d,
    _ => {},
}

File: examples/samples.rs [hunk 7]
```diff
@@ -0,0 +1,1175 @@
...
+            for i in 0..self.astar.len() {
+                let (x, y) = self.astar.get(i).unwrap();
+                console.set_char_background(x, y, self.light_ground, BackgroundFlag::Set);
+            }
+        } else {
+            iterate_map(&mut |x, y, c| {
+                let wall = c == '#';
+                if !wall {
+                    let d = self.dijkstra.distance_from_root((x, y));
+                    if d.is_some() {
```
Gustorn: `if let` is nicer.

examples/samples.rs
```diff
@@ -0,0 +1,1175 @@
...
+                        let color = colors::lerp(self.light_ground,
+                                                 self.dark_ground,
+                                                 0.9 * d.unwrap() / self.dijkstra_dist);
+                        console.set_char_background(x, y, color, BackgroundFlag::Set);
+                    }
+                }
+            });
+            for i in 0..self.dijkstra.len() {
+                let (x, y) = self.dijkstra.get(i).unwrap();
```
Gustorn: @tomassedovic this reminds me: should we implement `Index` on the paths? They basicly act like containers, and if the result of `get` is unwrapped anyway, we might as well provide a convenient way to get.

examples/samples.rs
```diff
@@ -0,0 +1,1175 @@
...
+    }
+
+    fn handle_event<F>(&mut self, console: &mut Offscreen, clo: &mut F)
+        where F: Fn(&mut PathSample) -> ()
+    {
+        console.put_char(self.dx, self.dy, self.old_char, BackgroundFlag::None);
+        clo(self);
+        self.old_char = console.get_char(self.dx, self.dy);
+        console.put_char(self.dx, self.dy, '+', BackgroundFlag::None);
+        if SMAP[self.dy as usize].to_string().chars().nth(self.dx as usize).unwrap() == ' ' {
```
Gustorn: You don't need the `to_string` part.

examples/samples.rs
```diff
@@ -0,0 +1,1175 @@
...
+
+        self.display_map(console);
+        self.draw_path(console);
+
+        self.busy -= system::get_last_frame_length();
+        if self.busy < 0.0 {
+            self.move_creature(console);
+        }
+
+        match event {
```
Gustorn: `if let` removes a level of nesting here too.

examples/samples.rs
```diff
@@ -0,0 +1,1175 @@
...
+            Some((_, Event::Key(state))) if state.key == Key::Special(KeyCode::Number1) => {
+                show_cursor(false);
+            },
+            Some((_, Event::Key(state))) if state.key == Key::Special(KeyCode::Number2) => {
+                show_cursor(true)
+            },
+            _ => {} // Ignore other events
+        }
+
+        if self.mouse_state.is_some() {
```
Gustorn:
```rust
if let Some(mouse) = self.mouse_state {
    console.print(1, 1, self.format(&mouse, root);
}
```

Gustorn:
My opinions can basicly be summarized as: you `if let` when applicable, it makes the code nicer. The only other thing (and I didn't want to spam the line comments with this) is inconsistent formatting: there are a lot of lines that are indented incorrectly.

examples/samples.rs
tomob: It would be also nice to iterate over a path:
```for (x, y) in path.iter() { ... }```

I was contemplating implementing it but finally decided to focus on porting alone.

**Comment by tomob** [comment]
1. Hmm, Emacs should have taken care of the tabs/spaces thing. I'll fix it.
2. `if let`. Right. I haven't internalised it yet :)

File: examples/samples.rs (hunk 8 above)
Comment (Gustorn) [inline examples/samples.rs]: Yeah, we already have an iterator, but that actually walks the path. It would be really easy to add an `iter` method for the path. I might as well implement it.

Comment (tomob) [inline examples/samples.rs]: Yes, I spent half an hour in total puzzlement trying to understand why my path disappears after I display it with `walk`. :)

Comment (Gustorn) [inline examples/samples.rs]: I implemented `path.iter()` (and tested it with your code: this samples example will make testing so much easier).

Comment (tomob) [inline examples/samples.rs]: The pleasure is mine.

**Comment by Gustorn** [comment]
Looking at the lot of `event` matches, they discard the `EventFlags` part is always discarded. Do you think `check_for_event ` should return some kind of new structure? So something like this would be possible:

```rust
// Current
match event {
    (_, Some(_)) => {},
    _ => {},
}

// Proposed:
match event.data {
    Some(_) => {},
    None => {},
}
```

This would make such matches much easier to read.

tomob:
This would simplify matching a lot. Do you also think of doing something like this?

```rust
match event.flags {
    KEY => { /* ... */ },
    MOUSE_MOVE => { /* ... */ },
    MOUSE_PRESS => { /* ... */ },
    _ => {}
}
```

Gustorn:
Yeah, I was thinking of splitting them up to `data` and `flags` so they can be matched separately.

examples/samples.rs
tomassedovic: @Gustorn yep, index on the paths makes a lot of sense to me. And sorry about the confusion with walk consuming the iterator. That's what libtcod does as well. But yeah, having a separate non-consuming iterator is a great idea.

tomassedovic:
Thanks a lot! This is a big chunk of code. Let's merge it now and we can improve it as we land things like the path iterator and keys reform (coming in an hour or so).

PR #186, opened 2015-08-27
Reference path iterator

Gustorn:
Implemented non-destructive path iterators, these should make displaying paths more convenient.

### DIFF
--- a/src/pathfinding.rs
+++ b/src/pathfinding.rs
@@ -88,7 +88,7 @@ impl<'a> AStar<'a> {
         }
     }
 
-    pub fn iter<'b>(&'b self) -> AStarPathIter<'a, 'b> {
+    pub fn iter(&'a self) -> AStarPathIter<'a> {
         AStarPathIter { current: -1, path: self }
     }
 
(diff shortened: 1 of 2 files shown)


src/pathfinding.rs
```diff
@@ -82,6 +88,10 @@ impl<'a> AStar<'a> {
         }
     }
 
+    pub fn iter<'b>(&'b self) -> AStarPathIter<'a, 'b> {
```
tomob: A Rust newbie question: How dow I read the lifetimes here? `AStarPathIter` object must live as long as both `'a` (i.e. `AStar` object) and `'b` (also an `AStar` object??)?

Gustorn: I actually overcomplicated it, but I'm going to answer here to avoid confusion after I push the new commit. It's best explained at the iterator struct definition:

```rust
pub struct DijkstraPathIter<'a, 'b> where 'a: 'b {
    current: i32,
    path: &'b Dijkstra<'a>,
}
```

This means that `Dijkstra` takes a lifetime parameter, that will live at least as long as the reference we're going to store.

tomob: I think I understand it. Thanks.

PR #187, opened 2015-08-27
Key reform

### LINKED ISSUE #184: Keyboard structs reform
I think I've screwed up the design of the keyboard structs. Observe what happens when you press the number `4` on the US keyboard layout alone and with `shift`:

    $ cargo run --example minimal
    Pressed key: KeyState { key: Special(Number4), pressed: true, left_alt: false, left_ctrl: false, right_alt: false, right_ctrl: false, shift: false }
    Pressed key: KeyState { key: Special(Number4), pressed: true, left_alt: false, left_ctrl: false, right_alt: false, right_ctrl: false, shift: true }

The latter case should give us some way of knowing that the `$` character was entered. There are cases (especially on the number row) where libtcod returns both a printable character and keycode.

So here's how we could fix it:

```rust
struct Key {
    code: KeyCode,
    printable: Option<char>,  // None if TCOD_key_t.c == 0
    left_alt, right_alt, left_ctrl, right_ctrl, shift, // unchanged
    alt: bool,  // proposed for easier matching: true if left_alt || right_alt
    ctrl: bool,  // proposed for easier matching: true if left_ctrl || right_ctrl
}
```

Questions:
1. Am I just missing something obvious that would not require the rewrite of every codebase using tcod?
2. Do we want `Option<char>` or just `char` with a potential `0` value? (the latter is what libtcod does)
3. alt/ctrl convenience fields: yay or nay? We could have them as methods instead, but that would make for a more verbose matching.

Here's how the new struct could be used:

```rust
use tcod::input::KeyCode::*;
match root.wait_for_keypress(false) {
    Key { code: Enter, alt: true, .. } => { // maximise the window }
    Key { printable: '$', .. } => { // pay for items in a shop }
    Key { printable: 'w', .. } | Key { code: NumPad5, .. } => { // wait a turn }
}
```

What do you folks think?

tomassedovic:
This fixes #184 because of which we couldn't access certain keys. The `KeyState` struct is renamed to `Key` and has fields `code` and `printable` that correspond to libtcod's `vk` and `c`.

This also adds two convenience fields: `alt` and `ctrl` that are true if the left or right variant is pressed.

This is a breaking change. To update your code, convert it to match on the `Key` struct. I.e. replace this:

```rust
match keypress.key {
    Special(Enter) if keypress.left_alt => { // fullscreen }
    Special(Escape) => { // exit game }
    Key::Printable('>') => { // descend }
}
```

with this:

```rust
match keypress {
    Key { code: Enter, alt: true, .. } => { // fullscreen }
    Key { code: Escape, .. } => { // exit game }
    Key { printable: '>', .. } => { // descend }
}
```

### DIFF
--- a/examples/keyboard.rs
+++ b/examples/keyboard.rs
@@ -1,7 +1,7 @@
 extern crate tcod;
 
 use tcod::{Console, RootConsole, BackgroundFlag};
-use tcod::input::Key::Special;
+use tcod::input::Key;
 use tcod::input::KeyCode::{Up, Down, Left, Right, Escape};
 
 fn main() {
@@ -20,12 +20,12 @@ fn main() {
         // libtcod 1.5.1 has a bug where `wait_for_keypress` emits two events:
         // one for key down and one for key up. So we ignore the "key up" ones.
         if keypress.pressed {
-            match keypress.key {
-                Special(Escape) => break,
-                Special(Up) => y -= 1,
-                Special(Down) => y += 1,
-                Special(Left) => x -= 1,
-                Special(Right) => x += 1,
+            match keypress {
+                Key { code: Escape, .. } => break,
+                Key { code: Up, .. } => y -= 1,
+                Key { code: Down, .. } => y += 1,
+                Key { code: Left, .. } => x -= 1,
+                Key { code: Right, .. } => x += 1,
                 _ => {}
             }
         }
--- a/examples/roguelike-tutorial-1.rs
+++ b/examples/roguelike-tutorial-1.rs
@@ -8,7 +8,7 @@ extern crate tcod;
 
 use tcod::console::{Root, Console, FontLayout, FontType, BackgroundFlag};
 use tcod::colors;
-use tcod::input::Key::Special;
+use tcod::input::Key;
 use tcod::input::KeyCode::{Up, Down, Left, Right, Escape, Enter};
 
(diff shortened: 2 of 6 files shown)


Gustorn:
I like it. My only complaint with the current design is that matching an event is ridiculous:

```rust
match event {
    Some(_, Event::Key(Key { code: Enter, .. })) => {},
}
```
That' s a lot of noise for nothing, but I don't think it's solvable in current Rust. Ideally I'd like to be able to do something like this:
```rust
match event { 
    Some(_, Event::Key { code: Enter, .. }) => {},
}
```

This could be done if we declared `Key` inside the `Event` struct, but unfortunately it's used in `console` as well (Which is the recommended method of getting events, `console` or `input`?).

tomassedovic:
Yeah, that is a bit silly. I wonder if the upcoming struct & enum unification could help here (haven't followed it too thoroughly yet).

If you're just getting keyboard input, console is the place to go, if you want mouse or both, then input. At least that's how I understand libtcod. It being a roguelike toolkit, mouse is sidelined a bit... Especially since it's more painful to match generic events instead of just the key ones.

Gustorn:
Well, other than that it seems good to me. Once it's merged I'll go and see if returning a dedicated `Event` type would make matching a bit easier (instead of the current tuple of (EventFlags, Event).

PR #188, opened 2015-08-28
Improvements

Gustorn:
I'm open to criticism regarding the changes in `samples`. I think matching agains either `key.code` or `key.printable` makes perfect sense: it improves readabilty (I don't have to look for `ifs` after the match). It can be inconsitent when you have to `match` for different event types though.


tomassedovic:
Neat! I'm +1 on all changes, but want give @tomob a chance to comment about the samples example cleanup.

tomob:
I'm ok with the cleanup. With changes to key events it looks really neat.

PR #191, opened 2015-08-30
Line toolkit

tomob:
Line sample – along with Line API.

The original API for drawing lines with a callback practically requires maintaining some global (static) state. This makes passing something reasonable, like a Rust closure, impossible. I'm contemplating adding a piece of C code to wrap line API with a C function that would take some user data, and in result allow for some tricks to pass rust closure or object.


Gustorn:
I'm a bit vary if the increasing use of `unsafe` and `transmute` in the samples code. If an itereable, wrapping enum is that widely used we should just create a safe wrapper somewhere in the beginning, and use that for all subsequent operations. This is not necessarily in the scope of this PR, but I think I'll do it once this is merged.

src/line.rs
```diff
@@ -0,0 +1,135 @@
...
+impl Line {
+    pub fn new(start: (i32, i32), end: (i32, i32)) -> Self {
+        let mut line: Line = Default::default();
+        unsafe {
+            ffi::TCOD_line_init_mt(start.0, start.1, end.0, end.1, &mut line.tcod_line)
+        };
+        line
+    }
+    
+    pub fn draw_with_callback(start: (i32, i32), end: (i32, i32), callback: Callback) -> Self {
```
Gustorn: This should be names `new_from_callback` if we're following the conventions in `pathfinding`

src/line.rs
```diff
@@ -0,0 +1,135 @@
...
+
+use bindings::ffi;
+use bindings::{c_int, c_bool};
+
+#[derive(Default)]
+pub struct Line {
+    tcod_line: ffi::TCOD_bresenham_data_t,
+}
+
+pub trait Listener {
```
Gustorn: I think this isn't used anywhere?

Gustorn:
As an additional note: isn't the whole line drawing toolkit a blob of mumtable global state? Because if that's the case we really shouldn't allow more than one instance to exist at any time.

src/line.rs
tomob: No, it's not. I wanted to create a nice object-base listener but couldn't make it work (as explained in the original pull request comment). I plan to continue working on it and I hope to make it work the way I want.

src/line.rs
tomob: Originally I called it `new_from_callback`, but calling the callback "consumes" the line, so the thing returned from the function is "the rest of the line". Hence the name.

src/line.rs
Gustorn: That's what I though, but I'd rather leave it out while it's not implemented.

tomob:
1. The background flag enum in libtcod is used both as enum (lowest byte) and alpha value (upper bytes). I really don't understand why they decided to code it this way (apart from ill-understood need to save space). I do agree that wrapping it in a nice Rust type (enum with values?) is a reasonable thing to do.
2. The line toolkit has two sets of functions: one with `_mt` suffix (presumably "multithreaded") that passes line data to all calls, and one without the suffix which relies on global state. I decided to use the former, therefore there can be more that one line at the same time.

src/line.rs
tomob: OK, I'll update the pull request tomorrow (or, actually, later today, when I catch some sleep :) ).

Gustorn:
Sorry about the 2nd point, I was mildly sleep-deprived when making it: the `_mt` versions are actually safe (they just aren't documented).

src/line.rs
tomassedovic: What about `new_with_callback`? Not particularly against the current name but I would mildly prefer having `new` in there somewhere.

src/line.rs
```diff
@@ -0,0 +1,131 @@
...
+
+use bindings::ffi;
+use bindings::{c_int, c_bool};
+
+#[derive(Default)]
+pub struct Line {
+    tcod_line: ffi::TCOD_bresenham_data_t,
+}
+
+pub type Callback = extern "C" fn(x: c_int, y: c_int) -> c_bool;
```
tomassedovic: I would vastly prefer a closure here but I'm not sure it's even possible. The C callback doesn't accept user_data, unfortunately. Not sure we can even make it use `i32` and `bool` types.

So yeah, this is fine but if anyone has any idea how to make this more usable, I'm all ears.

tomassedovic:
Okay, I'm fine merging this as is but if we're all happy about `Line::new_with_callback`, I'd like that better.

Not to happy about the callback situation, but there probably isn't much we can do about that. Agreed re: unsafe but that can happen in another PR.

Gustorn:
It wouldn't be that hard to accept Rust closures. This is the `C` function that would have to be ported:

```c
bool TCOD_line_mt(int xo, int yo, int xd, int yd, TCOD_line_listener_t listener, TCOD_bresenham_data_t *data) {
	TCOD_line_init_mt(xo,yo,xd,yd,data);
	do {
		if (! listener(xo,yo)) return false;
	} while (! TCOD_line_step_mt(&xo,&yo,data));
	return true;
}
```

src/line.rs
tomob: `new_with_callback` it is. Will push a change today.

tomob:
I'm working on the callback thing, but I'd rather have it as a separate pull request, because it will take me a couple of days, probably (I'm quite busy at the moment).

Gustorn:
I don't think closures can be done more elegantly than this: [patch](https://gist.github.com/Gustorn/3878f29afe904006fb05). You either do this or submit a patch upstream to accept user data...

tomassedovic:
I definitely don't want to block this on the callback work. Let's merge this after the rename.

tomob:
The callbacks were simpler then I expected. Here you are.

tomassedovic:
Cool! One question though: isn't `Listener` redundant? It seems to me you could get the same functionality by implementing the `FnMut(i32, i32) -> bool` trait on your custom struct (and using the callback code).

Sorry, this is probably a really silly question but it's getting late and I'm no longer thinking clear.

tomob:
You're right. It seems I'm following C++ examples much too closely. :)

tomassedovic:
Ah okay :-). Can you remove that real quick then so I can merge this clean? Or we can do that in a followup pr...

tomob:
Hm... But can Fn* be implemented in stable Rust?

Gustorn:
No, but you don't need to. The user can still pass functions or closures.

tomassedovic:
I think @tomob means for the users? As in tcod itself doesn't need to but if you want to use a custom struct as a callback, you'll have to use unsafe rust. Not sure that bothers me, but I did assume you could just impl Fn* in stable...

Gustorn:
You cannot: [docs](https://doc.rust-lang.org/stable/std/ops/trait.Fn.html). It still needs a bit of ugly syntax, even in nightly:

```rust
impl FnOnce<(i32,i32)> for Type {
    type Output = bool;

    extern "rust-call" fn call_once(self, args: (i32, i32)) -> Self::Output { 
        //
    }
}
```

Maybe test if we could blanket impl `Listener` to all `Fn` types? That way closures still work but users could still define their own structs.

tomob:
I'm not that interested in having a listener API. I followed what C++ bindings do (at least on interface level), but a listener can be trivially implemented with a closure:
```rust
let listener = MyListener::new(whatever);
line.new_with_callback((x0, y0), (x1, y1), |x, y| listener.put_point(x, y));
```
So I don't think it's a big deal.

tomassedovic:
All right, thanks a lot! Sorry it took so long to merge, but I think the code & api are better for it.

PR #193, opened 2015-08-31
Safe iterable/wrapping enums

Gustorn:
An example for safe iterable/wrapping enum implementation: I realize that providing this functionality as part of the library is questionable, but there is a problem with providing this implementation in the examples (or leaving it to the users): the `mem::transmute(enum as i32 + 1)` trick is not guaranteed to last forever, upstream could decide (for some reason) to assign actual values to these enums. This version will at least catch the error if you run `cargo test`.

The other question is: is this really necessary? It sure as hell is nice with the `samples` code (especially once #191 is merged), but I don't really see this as something I would use every day.

### DIFF
--- a/examples/samples.rs
+++ b/examples/samples.rs
@@ -320,20 +320,6 @@ impl FovSample {
         console.put_char(self.px, self.py, '@', BackgroundFlag::None);
         self.recompute_fov = true;
     }
-
-    fn next_algorithm(&mut self) {
-        match self.algorithm {
-            FovAlgorithm::Restrictive => return,
-            _ => self.algorithm = unsafe { std::mem::transmute(self.algorithm as i32 + 1) }
-        };
-    }
-
-    fn previous_algorithm(&mut self) {
-        match self.algorithm {
-            FovAlgorithm::Basic => return,
-            _ => self.algorithm = unsafe { std::mem::transmute(self.algorithm as i32 - 1) }
-        };
-    }
 }
 
 impl Render for FovSample {
@@ -392,12 +378,12 @@ impl Render for FovSample {
                     self.recompute_fov = true;
                 },
                 '+' => {
-                    self.next_algorithm();
+                    self.algorithm = self.algorithm.bounded_next();
                     self.display_help(console);
                     self.recompute_fov = true;
                 },
                 '-' => {
-                    self.previous_algorithm();
+                    self.algorithm = self.algorithm.bounded_prev();
                     self.display_help(console);
                     self.recompute_fov = true;
                 },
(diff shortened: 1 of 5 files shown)


Gustorn:
So, I was wondering why this failed, since according to the current docs, I'm not using anything that's unstable. The problem is, we're still building under 1.0 when 1.2 stable is out. Shouldn't we change the explicit `1.0` to `stable`?

tomassedovic:
Basically, I didn't want to accidentally require a newer version. This seems like a good-enough reason so sure. I'll update travis.yml. Also, there's an option to migrate to a container-based setup so I'll look into that, too.

tomassedovic:
Regarding the actual change: I don't want our users to use `unsafe` with tcod -- especially `transmute` -- because it's _nicer_. Sure, if they need raw libtcod bindings because of performance, no problem.

Being able to iterate and rotate enums is a general thing that I think is quite common outside of tcod (I hit it and so do people on reddit regularly) and there's also this: https://crates.io/crates/enum_derive although it works a bit differently.

But yeah, I've no trouble merging this and switching to something better once there is a proper way to do this in the wider ecosystem (could be stable custom plugins, iterable enums in the core lang or whatever).

Gustorn:
`enum_derive` sounds cool (and I think it works on `stable`?), but it doesn't provide:
1. The wrapping behaviour
2. Stepping in arbitrary directions (`next`, `prev`).
That the `samples` code needs. I don't really know if those two things are really useful outside of this specific usecase though, if they aren't it might be a better idea to just use a library that's already there (`enum_derive`).

tomassedovic:
I was thinking something like enum_iter.cycle() would give you an infinite iterator. But it doesn't implement DoubleEndedIterator so you wouldn't be able to go back.

Let's just go with this for now.

tomassedovic:
Okay, travis should be tracking stable now. Also OSX :-)

tomassedovic:
I keep going back and forth on this. It would make the samples code simpler, but you're right, I don't see cycling between background flags or available renderers a common thing to do either.

I'm still hoping that we'll get proper iteration and prev/next cycling of enums in Rust or at least through a common plugin, but yeah, maybe baking it here is not the ideal solution since it's probably usable just in the samples example.

I wonder if we could implement cycles and iteration directly there. Ideally with a full match rather than transmute. A macro could help us simplify that.

Gustorn:
I'd actually vote for not merging this, even though I'm the one who implemented it. This doesn't seem like a feature `libtcod` should provide.

tomassedovic:
Yeah agreed. Let's see if we can make samples nicer directly instead.

PR #196, opened 2015-09-02
Wrapper around Noise

tomob:
Along with demo in samples.

### DIFF
--- a/examples/samples.rs
+++ b/examples/samples.rs
@@ -13,6 +13,7 @@ use tcod::map::{Map, FovAlgorithm};
 use tcod::image;
 use tcod::namegen::Namegen;
 use tcod::line::Line;
+use tcod::noise::Noise;
 use rand::Rng;
 use rand::ThreadRng;
 use std::char::from_u32;
@@ -351,10 +352,10 @@ struct FovSample {
     light_wall: colors::Color,
     dark_ground: colors::Color,
     light_ground: colors::Color,
-    // noise: Noise,
+    noise: Noise,
     light_walls: bool,
     algorithm: FovAlgorithm,
-    _torch_x: f32,
+    torch_x: f32,
 }
 
 fn clamp(a: f32, b: f32, x: f32) -> f32 {
@@ -372,10 +373,10 @@ impl FovSample {
             light_wall: colors::Color::new(130, 110, 50),
             dark_ground: colors::Color::new(50, 50, 150),
             light_ground: colors::Color::new(200, 180, 50),
-            // noise: ,
+            noise: Noise::initializer().dimensions(1).init(),
             light_walls: true,
             algorithm: FovAlgorithm::Basic,
-            _torch_x: 0.0,
+            torch_x: 0.0,
         }
     }
 
@@ -466,11 +467,16 @@ impl Render for FovSample {
             self.map.compute_fov(self.px, self.py, radius, self.light_walls, self.algorithm);
         }
 
-        let dx = 0.0;
-        let dy = 0.0;
-        let di = 0.0;
+        let mut dx = 0.0;
+        let mut dy = 0.0;
+        let mut di = 0.0;
         if self.torch {
-            // TODO implemenent when noise is wrapped in Rust API
+            self.torch_x += 0.2;
+
+            let tdx = self.torch_x + 20.0;
+            dx = self.noise.get(&mut [tdx]) * 1.5;
+            dy = self.noise.get(&mut [tdx + 30.0]) * 1.5;
+            di = self.noise.get(&mut [self.torch_x]) * 0.2;
         }
 
         self.display_map(console, dx, dy, di);
--- a/examples/samples.rs
+++ b/examples/samples.rs
@@ -408,20 +408,8 @@ impl NoiseSample {
             .lacunarity(self.lacunarity)
             .init()
     }
-}
 
-impl Render for NoiseSample {
-    fn initialize(&mut self, _console: &mut Offscreen) {
-        system::set_fps(30);
-    }
-
-    fn render(&mut self,
-              console: &mut Offscreen,
-              _root: &Root,
-              event: Option<(EventFlags, Event)>) {
-        self.dx += 0.01;
-        self.dy += 0.01;
-
+    fn draw_noise(&mut self) {
         for y in 0..2*SAMPLE_SCREEN_HEIGHT {
             for x in 0..2*SAMPLE_SCREEN_WIDTH {
                 let x0 = self.zoom * x as f32 / (2 * SAMPLE_SCREEN_WIDTH) as f32 + self.dx;
@@ -453,8 +441,9 @@ impl Render for NoiseSample {
                 self.img.put_pixel(x, y, color);
             }
         }
+    }
 
-        blit_2x(&self.img, (0, 0), (-1, -1), console, (0, 0));
+    fn draw_rectangle(&self, console: &mut Offscreen) {
         console.set_default_background(colors::GREY);
         let height = if self.func as u32 <= NoiseType::Wavelet as u32 {10} else {13};
         console.rect(2, 2, 23, height, false, BackgroundFlag::Multiply);
@@ -465,7 +454,9 @@ impl Render for NoiseSample {
                 console.set_char_foreground(x, y, color);
             }
         }
+    }
 
+    fn draw_menu(&self, console: &mut Offscreen) {
         for cur_func in (NoiseFunction::Perlin as i32)..(NoiseFunction::TurbulenceWavelet as i32 + 1) {
             if self.func as i32 == cur_func {
                 console.set_default_foreground(colors::WHITE);
@@ -485,6 +476,27 @@ impl Render for NoiseSample {
             console.print(2, 13, format!("R/F : lacunarity ({:2.1})", self.lacunarity));
             console.print(2, 14, format!("T/G : octaves ({})", self.octaves));
         }
+
+    }
+}
+
+impl Render for NoiseSample {
+    fn initialize(&mut self, _console: &mut Offscreen) {
+        system::set_fps(30);
+    }
+
+    fn render(&mut self,
+              console: &mut Offscreen,
+              _root: &Root,
+              event: Option<(EventFlags, Event)>) {
+        self.dx += 0.01;
+        self.dy += 0.01;
+
+        self.draw_noise();
+        blit_2x(&self.img, (0, 0), (-1, -1), console, (0, 0));
+
+        self.draw_rectangle(console);
+        self.draw_menu(console);
 
         if let Some((_, Event::Key(key))) = event {
             match key.printable {
--- a/examples/samples.rs
+++ b/examples/samples.rs
@@ -343,8 +343,7 @@ impl Render for LineSample {
     }
 }
 
-#[allow(dead_code)]
-#[derive(Clone,Copy,Debug)]
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
 enum NoiseFunction {
     Perlin = 0,
     Simplex,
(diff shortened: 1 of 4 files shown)


examples/samples.rs
```diff
@@ -341,6 +343,195 @@ impl Render for LineSample {
     }
 }
 
+#[allow(dead_code)]
+#[derive(Clone,Copy,Debug)]
+enum NoiseFunction {
+    Perlin = 0,
+    Simplex,
+    Wavelet,
+	FbmPerlin,
```
Gustorn: I think some of your Tab->Space conversions have gone wrong.

src/noise.rs
```diff
@@ -0,0 +1,454 @@
...
+extern crate libc;
+
+use bindings::ffi;
+use bindings::AsNative;
+
+use random::Rng;
+
+/// Available noise types.
+#[repr(C)]
+#[derive(Copy, Clone)]
```
Gustorn: Could you also derive `PartialEq, Eq and Debug`? When writing the safe enum wrapping code I found that they'll be needed sooner or later.

src/noise.rs
```diff
@@ -0,0 +1,454 @@
+//! Noise toolkit.
+//!
+//! Rust bindings follow the original API pretty closely.
+
+extern crate libc;
```
Gustorn: I don't think this `extern crate` is needed.

src/noise.rs
```diff
@@ -0,0 +1,454 @@
...
+pub struct NoiseInitializer {
+    dimensions: u32,
+    hurst: f32,
+    lacunarity: f32,
+    noise_type: NoiseType,
+    random: Rng
+}
+
+impl NoiseInitializer {
+    fn new() -> Self {
```
Gustorn: This is just an opinion, but I think dimensions should be always explicitly specified. The other defaults make perfect sense.

examples/samples.rs
```diff
@@ -341,6 +343,195 @@ impl Render for LineSample {
...
+        self.draw_noise();
+        blit_2x(&self.img, (0, 0), (-1, -1), console, (0, 0));
+
+        self.draw_rectangle(console);
+        self.draw_menu(console);
+
+        if let Some((_, Event::Key(key))) = event {
+            match key.printable {
+                '1'...'9' =>
+                    self.func = unsafe{ std::mem::transmute(key.printable as u8 - '1' as u8) },
```
Gustorn: Wouldn't [`to_digit`](https://doc.rust-lang.org/stable/std/primitive.char.html#method.to_digit) work here?

examples/samples.rs
tomob: Yeah, copy-pasting from C++ code :/

src/noise.rs
tomob: Indeed it isn't.

examples/samples.rs
tomob: Yes. I don't know the Rust std lib well enough yet :)

src/noise.rs
tomob: What is the idiomatic way to implement it? I'm thinking of setting dimensions to 0 in `new` and calling `panic!` in `init` if it's still zero.

(Actually I should `assert` then dimensions is between 1 and 4.)

Gustorn: I would just replace `new()` with `with_dimensions(dimensions: u32)`.

tomob: Aah, that's simple and fool-proof!

examples/samples.rs
```diff
@@ -341,6 +343,196 @@ impl Render for LineSample {
...
+            for x in 2..25 {
+                let old_col = console.get_char_foreground(x, y);
+                let color = old_col * colors::GREY;
+                console.set_char_foreground(x, y, color);
+            }
+        }
+    }
+
+    fn draw_menu(&self, console: &mut Offscreen) {
+        for cur_func in (NoiseFunction::Perlin as i32)..(NoiseFunction::TurbulenceWavelet as i32 + 1) {
```
Gustorn: I'm not fond of this, but I concede that there is no better way as of yet. Rust really needs some iterable enums... Other than this I'm fine with the merge (and great job with wrapping two of the bigger remaining `libtcod` APIs).

tomob: Yes, this hurts my eyes when I look at it but I found no elegant way to do it.

To complete my porting of `samples.cpp` I need to wrap BSP toolkit. And when I'm done with samples, I'll get to implementing a game. :)

tomob: Well, how about this?

```rust
struct NoiseFunctionIterator {
    val: u8
}

impl NoiseFunction {
    fn iter() -> NoiseFunctionIterator {
        NoiseFunctionIterator { val: 0 }
    }
}

impl Iterator for NoiseFunctionIterator {
    type Item = NoiseFunction;
    
    fn next(&mut self) -> Option<Self::Item> {
        match self.val {
            0...8 => {
                let retval = unsafe { std::mem::transmute(self.val) };
                self.val += 1;
                Some(retval)
            },
            _ => None,
        }
    }
}

// ...
for e in NoiseFunction::iter() {
    //...
}
```

Gustorn:
That's much better. You could make a static array in `next` so that you don't depend on the underlying type being in `0...8`. Example (untested, but should work):

```rust
impl Iterator for NoiseFunctionIterator {
    type Item = NoiseFunction;

    fn next(&mut self) -> Option<Self::Item> {
        use self::NoiseFunction::*;
        static VALUES: &'static [NoiseFunction] = &[A, B, C, D];
        match self.val {
            0...8 => {
                let ret = VALUES[self.val as usize];
                self.val += 1;
                Some(ret)
            },
            _ => None,
        }
    }
}
```

It also generates good assembly: [godbolt](https://goo.gl/tlupyk)

examples/samples.rs
```diff
@@ -341,6 +343,230 @@ impl Render for LineSample {
...
+            dy: 0.0,
+            octaves: 4,
+            hurst: DEFAULT_HURST,
+            lacunarity: DEFAULT_LACUNARITY,
+            img: Image::new(SAMPLE_SCREEN_WIDTH * 2, SAMPLE_SCREEN_HEIGHT * 2),
+            zoom: 3.0
+        }
+    }
+
+    fn new_noice(&self) -> Noise {
```
Gustorn: `noice` -> `noise`

examples/samples.rs
```diff
@@ -341,6 +343,230 @@ impl Render for LineSample {
...
+
+        self.draw_rectangle(console);
+        self.draw_menu(console);
+
+        if let Some((_, Event::Key(key))) = event {
+            match key.printable {
+                '1'...'9' =>
+                    self.func = unsafe{
+                        let number = key.printable.to_digit(10).unwrap() as u8;
+                        std::mem::transmute(number - 1)
```
Gustorn: You don't need transmute here. Just subtract `1` and use `as` to cast to the correct type. Using transmute when there is an alternative solution is generally a bad idea.

src/noise.rs
```diff
@@ -0,0 +1,452 @@
...
+pub struct NoiseInitializer {
+    dimensions: u32,
+    hurst: f32,
+    lacunarity: f32,
+    noise_type: NoiseType,
+    random: Rng
+}
+
+impl NoiseInitializer {
+    fn new_init_with_dimensions(dimensions: u32) -> Self {
```
Gustorn: `assert!` on dimensions being in the correct range?

src/noise.rs
```diff
@@ -0,0 +1,452 @@
...
+        NoiseInitializer {
+            dimensions: dimensions,
+            hurst: DEFAULT_HURST,
+            lacunarity: DEFAULT_LACUNARITY,
+            noise_type: NoiseType::Default,
+            random: Rng::get_instance()
+        }
+    }
+
+    /// Sets the dimensions of the noise generator. Defaults to 2.
```
Gustorn: Doesn't default to anything anymore.

examples/samples.rs
tomob: I'm making this mistake EVERY F*G TIME I write this word.

examples/samples.rs
tomob: Umm... No?

```
examples/samples.rs:538:25: 538:54 error: non-scalar cast: `u8` as `NoiseFunction`
examples/samples.rs:538                         (number - 1) as NoiseFunction
                                                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

src/noise.rs
tomob: Right, I meant to do it.

examples/samples.rs
Gustorn: Yeah, you're right. I'd still prefer an impl on `NoiseFunction`:

```rust
impl NoiseFunction {
    fn from_value(val: u8) -> Self {
        // Do the same static array indexing trick as in Iterator
    }
}
```

src/noise.rs
```diff
@@ -0,0 +1,452 @@
...
+        NoiseInitializer::new_init_with_dimensions(dimensions)
+    }
+
+    pub fn set_type(&self, noise_type: NoiseType) {
+        unsafe {
+            ffi::TCOD_noise_set_type(self.noise, noise_type as u32)
+        }
+    }
+
+    pub fn get(&self, coords: &mut [f32]) -> f32 {
```
tomassedovic: Seems to me coords here (and in the other get methods) should be an immutable array, but tcod_sys requires a mut pointer right now. I think we can change that to be a `const *`. But that should happen separately from this.

tomob: Do you mean a change in tcod_sys?

tomassedovic: On the other hand, what if we accepted the coords array by value? I think it would make for a slightly nicer API here and since the max size is 4 the overhead should be pretty small.

tomassedovic: Yeah, in tcod_sys. But I don't want to block this change on that.

Gustorn: If it's by value, then just accept `T: AsMut<[T]>` for now, and investigate changing the coords to `const*`.

Gustorn: The problem with the `tcod_sys` change is that we'd have to change the signiture of the `C` functions to be completely safe. Just casting `*const` to `*mut` seems like a bad idea.

tomob: \<newbie_mode>
Like this?

```rust
    pub fn get<T: AsMut<[f32]>>(&self, mut coords: T) -> f32 {
        assert!(self.dimensions as usize == coords.as_mut().len());
        unsafe {
            ffi::TCOD_noise_get(self.noise, coords.as_mut().as_mut_ptr())
        }
    }
```
\<\newbie_mode>

Gustorn: Yes, that should work fine.

tomassedovic: @tomob yep.

@Gustorn hm, isn't it a fairly standard thing to do with C bindings in Rust? I'll look at other projects but I think that enforcing const-ness on the tcod_sys level is enough (provided C does not mutate the args). We could always just clone the array to be completely safe but that seems like an unnecessary overhead.

Gustorn: @tomassedovic I don't mind changing `tcod_sys`, and we can prove that the current version included with `tcod-rs` will not mutate `coords`. My problem is that we support other precompiled versions: if any of those (for some reason) change this behaviour things will silently fail. It's not terribly likely to happen, so I'm fine with doing it, but it's something to consider.

tomassedovic: Yeah, fair enough. I'm quite happy with the AsMut interface anyway, so we may just leave that as is.

tomassedovic:
:+1: from me.

@Gustorn you happy to merge this?

src/noise.rs
```diff
@@ -0,0 +1,469 @@
...
+            dimensions: dimensions,
+            hurst: DEFAULT_HURST,
+            lacunarity: DEFAULT_LACUNARITY,
+            noise_type: NoiseType::Default,
+            random: Rng::get_instance()
+        }
+    }
+
+    /// Sets the dimensions of the noise generator.
+    pub fn dimensions(&mut self, dimensions: u32) -> &mut Self {
```
Gustorn: One more thing: now that I think about it, this function is completely redundant, right? We're required to set our dimensions at construction, it doesn't really make any sense to change it after that.

tomob: Removed.

Gustorn:
@tomassedovic I'm fine with merging this now.

tomassedovic:
Thanks a lot!

tomob:
Thank you, guys.

tomassedovic:
@tomob Thank *you* for writing the code and putting up with our comments!

PR #198, opened 2015-09-10
BSP toolkit

tomob:
BSP toolkit and related example.

### DIFF
--- a/examples/samples.rs
+++ b/examples/samples.rs
@@ -15,6 +15,7 @@ use tcod::namegen::Namegen;
 use tcod::line::Line;
 use tcod::noise::{Noise, NoiseType, DEFAULT_HURST, DEFAULT_LACUNARITY, MAX_OCTAVES};
 use tcod::image::{Image, blit_2x};
+use tcod::bsp::BSP;
 use rand::Rng;
 use rand::ThreadRng;
 use std::char::from_u32;
@@ -994,6 +995,46 @@ impl<'a> Render for PathSample<'a> {
     }
 }
 
+struct BspListener;
+
+impl BspListener {
+    fn node(bsp: &BSP) {
+    }
+}
+
+struct BspSample {
+    bsp: BSP,
+    generate: bool,
+    refresh: bool,
+    map: Map,
+    dark_wall: colors::Color,
+    dark_ground: colors::Color,
+    listener: BspListener,
+}
+
+impl BspSample {
+    fn new() -> Self {
+        BspSample {
+            bsp: BSP::new_with_size(0, 0, SAMPLE_SCREEN_WIDTH, SAMPLE_SCREEN_HEIGHT),
+            generate: true,
+            refresh: false,
+            map: Map::new(SAMPLE_SCREEN_WIDTH, SAMPLE_SCREEN_HEIGHT),
+            dark_wall: colors::Color::new(0, 0, 100),
+            dark_ground: colors::Color::new(50, 50, 150),
+            listener: BspListener
+        }
+    }
+}
+
+impl Render for BspSample {
+    fn initialize(&mut self, console: &mut Offscreen) {
+    }
+
+    fn render(&mut self, console: &mut Offscreen, root: &Root,
+              event: Option<(EventFlags, Event)>) {
+    }
+}
+
 struct ImageSample {
     img: image::Image,
     circle: image::Image,
@@ -1301,13 +1342,14 @@ fn main() {
     let mut names = NameSample::new();
     let mut line = LineSample::new();
     let mut noise = NoiseSample::new();
+    let mut bsp = BspSample::new();
     let mut samples = vec![MenuItem::new("  True colors      ", &mut colors),
                            MenuItem::new("  Offscreen console", &mut offscreen),
                            MenuItem::new("  Line drawing     ", &mut line),
                            MenuItem::new("  Noise            ", &mut noise),
                            MenuItem::new("  Field of view    ", &mut fov),
                            MenuItem::new("  Path finding     ", &mut path_sample),
-                           // MenuItem::new("  Bsp toolkit      ", &mut ),
+                           MenuItem::new("  Bsp toolkit      ", &mut bsp),
                            MenuItem::new("  Image toolkit    ", &mut image_sample),
                            MenuItem::new("  Mouse support    ", &mut mouse),
                            MenuItem::new("  Name generator   ", &mut names),
--- a/examples/samples.rs
+++ b/examples/samples.rs
@@ -15,7 +15,7 @@ use tcod::namegen::Namegen;
 use tcod::line::Line;
 use tcod::noise::{Noise, NoiseType, DEFAULT_HURST, DEFAULT_LACUNARITY, MAX_OCTAVES};
 use tcod::image::{Image, blit_2x};
-use tcod::bsp::{BSP, TraverseOrder};
+use tcod::bsp::{Bsp, TraverseOrder};
 use rand::Rng;
 use rand::ThreadRng;
 use std::char::from_u32;
@@ -999,7 +999,7 @@ impl<'a> Render for PathSample<'a> {
 type CharMap = [[char; SAMPLE_SCREEN_WIDTH as usize]; SAMPLE_SCREEN_HEIGHT as usize];
 
 struct BspSample {
-    bsp: BSP,
+    bsp: Bsp,
     generate: bool,
     refresh: bool,
     map: CharMap,
@@ -1021,7 +1021,7 @@ fn random_val(mut low: i32, high: i32) -> i32 {
 impl BspSample {
     fn new() -> Self {
         BspSample {
-            bsp: BSP::new_with_size(0, 0, SAMPLE_SCREEN_WIDTH, SAMPLE_SCREEN_HEIGHT),
+            bsp: Bsp::new_with_size(0, 0, SAMPLE_SCREEN_WIDTH, SAMPLE_SCREEN_HEIGHT),
             generate: true,
             refresh: false,
             map: [['#'; SAMPLE_SCREEN_WIDTH as usize]; SAMPLE_SCREEN_HEIGHT as usize],
@@ -1058,7 +1058,7 @@ impl BspSample {
         self.refresh  = false;
     }
 
-    fn visit(&self, node: &mut BSP, map: &mut CharMap) -> bool {
+    fn visit(&self, node: &mut Bsp, map: &mut CharMap) -> bool {
         if node.is_leaf() {
             self.visit_leaf(node, map);
         } else {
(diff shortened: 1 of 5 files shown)


examples/samples.rs
```diff
@@ -994,6 +996,276 @@ impl<'a> Render for PathSample<'a> {
...
+            min_room_size: 4,
+            random_room: false,
+            room_walls: true,
+        }
+    }
+
+    fn rebuild(&mut self) {
+        self.bsp.resize(0, 0, SAMPLE_SCREEN_WIDTH, SAMPLE_SCREEN_HEIGHT);
+        for row in &mut self.map {
+            for item in &mut row[..] {
```
Gustorn: You shouldn't need the slicing syntax here (`[..]`). If you are on the nightly, building once with `clippy` will help you catch these stylistic suggestions (don't forget to remove it from the final version).

Gustorn: Never mind, the array is too big. I hate not having type level integers.

src/bsp.rs
```diff
@@ -0,0 +1,416 @@
...
+
+impl Deref for BSP {
+    type Target = ffi::TCOD_bsp_t;
+
+    fn deref(&self) -> &Self::Target {
+        unsafe { &*self.bsp }
+    }
+}
+
+impl DerefMut for BSP {
```
Gustorn: I'm incredibly uncomfortable with an implicit deref into a native type. I'd rather limit the casting to/from native types to `AsNative` and `FromNative`.

src/bsp.rs
```diff
@@ -0,0 +1,416 @@
...
+/// assert_eq!(bsp.y, 0);
+/// assert_eq!(bsp.w, 50);
+/// assert_eq!(bsp.h, 60);
+/// assert_eq!(bsp.horizontal(), false);
+///
+/// bsp.x = 10;
+/// bsp.y = 20;
+/// bsp.set_horizontal(true);
+/// ```
+pub struct BSP {
```
Gustorn: I think this should be named `Bsp` if we're following the Rust conventions.

examples/samples.rs
tomob: I'm on Rust 1.2, I am still to figure out how to work with multirust.

I could use `for item in &mut row.iter_mut()`. I don't remember why I used slicing. Which one would be better in this case (and why)?

src/bsp.rs
tomob: Yet, doing things like `bsp.x = 10` is very elegant, isn't it? There's no way to implement a getter/setter for a field, anything like Python's `@property`. Or is it?

However, it contains a bug. There's a possibility that `bsp` field is NULL. Let me fix it...

src/bsp.rs
tomob: :/
Right. Let me do some replacing...

examples/samples.rs
Gustorn: They should be equivalent. Both of them will yield a `std::slice::IterMut` in the end.

src/bsp.rs
```diff
@@ -0,0 +1,434 @@
...
+/// bsp.x = 10;
+/// bsp.y = 20;
+/// bsp.set_horizontal(true);
+/// ```
+pub struct Bsp {
+    bsp: *mut ffi::TCOD_bsp_t,
+    root: bool
+}
+
+impl Deref for Bsp {
```
Gustorn: I still dislike the auto deref. You could just reproduce the C struct in Rust, make it `repr(C)` and transmute into it. You can then expose everything except for the unsafe pointers.

tomob: I was offline for the weekend. I'll get to it over the next couple of days (unless @tomassedovic says it's OK :) )

examples/samples.rs
tomob: I think `iter_mut` reflect intentions better than slicing.

src/bsp.rs
tomassedovic: Sorry, didn't have time earlier. I'm not too happy about the autoderef here either. We have the `AsNative` trait. I think it can be used for this as well:

https://github.com/tomassedovic/tcod-rs/blob/master/src/bindings.rs#L13

Looks great otherwise so if we can drop the deref impl, I'm happy to merge this.

tomob: I wouldn't want to force users to call `as_native` when they want to read properties of the tree. I can try @Gustorn's approach or define accessors, as with `horizontal`.

Gustorn: I didn't mean accessors. Just define a struct: 

```rust
#[repr(C)]
struct Bsp {
    tree: TCOD_bsp_t,    
    // All other Bsp members made public, matching the declaration order of the C struct
}
```

Then you can just create an instance of the C struct an directly transmute it into the Rust version. It should be a free operation and hides the ugly unsafe pointer in the Rust API.

tomob:
I got rid of the raw pointer in favor of a Rust reference (and a couple of checks to make sure it's always valid).

Gustorn:
I'm not sure why the pointer -> ref conversion was necessary. My problem wasn't the `unsafe` block in `Deref`, it's that `bsp.tree` exposes a raw pointer in a safe Rust API.

tomob:
The main problem is those pointers are the only thing libtcod exposes. It allocates structs internally (for root node and all children) and deletes them, while allowing users of the toolkit to read *and write* to all nodes.

I tried several approaches and it either causes problems with deallocating nodes or introduces a lot of copying contents TCOD_bsd_t there and back.

I could implement accessors (though I consider derefing a much cleaner API for users) and this way I would avoid ever showing the pointer to anyone :).

Gustorn:
I'm honestly curious what's the problem with this approach (I didn't have the time to evaluate it, but if you did, please share):

```rust
#[repr(C)]
pub struct BspInner {
    tree: TCOD_tree_t,
    pub x: i32,
    pub y: i32,
    pub w: i32,
    pub h: i32,
    pub position: i32,
    pub level: u8,
    pub horizontal: bool,
}

impl Bsp {
    pub fn new_with_size(x: i32, y: i32, w: i32, h: i32) -> Self {
        let bsp = unsafe {
            let pointer = ffi::TCOD_bsp_new_with_size(x, y, w, h);
            if pointer.is_null() {
                panic!("TCOD_bsp_new_with_size returned a NULL BSP.");
            }
            let mut t = mem::transmute::<ffi::TCOD_bsp_t, BspInner>(*pointer);
            t.horizontal = (*pointer).horizontal != 0; // C and Rust bool semantics don't match exactly
            t
        };
        Bsp { bsp: bsp, root: true }
    }
}
```

Sorry if this has syntax errors, but I don't have access to a Rust compiler right now.

tomob:
@Gustorn Heh, I missed your edit and started writing a lengthy comment with GDB session excerpts to illustrate my point. :)

@tomassedovic What's your take on this? We still have three options: 1. leaving it as is (no more work, potentially unsafe), 2. replacing derefs with accessors (safe and requiring relatively small amount of work), and 3. rewriting it all in Rust (safest, some work, requires tracking changes in libtcod until heat death of the universe).

After spending on it unreasonable amount of time and implementing half a dozen of approaches (not shown here), I would personally lean towards 2: even though I think it's the ugliest, it's also the safest use of the underlying C struct.

Gustorn:
I really wanted to rewrite `namegen` and `random` in Rust too, but I refrained from doing so to keep complete compatibility (we cannot really guarantee that we can match the original behaviour, especially if `libtcod` changes something under the hood). All in all I don't think that's a viable option.

I really hate trivial accessors, but they are probably the safest solution. The only thing is that I actually couldn't produce any unsafe behaviour with the exposed tree pointer. If someone could produce a null deref or a memory leak then the choice would be pretty straightforward (go for the accessors), but until that happens I'm on the fence: I very much prefer `bsp.x` to `bsp.x()`.

tomassedovic:
Looking. Sorry again, kind of a busy week.

tomassedovic:
Isn't the `bsp.tree` pointer always a raw one? I.e. the only way to mess something up is through an `unsafe` block? If that's the case I'm fine leaving it as is.

Regarding rewriting parts of libtcod in Rust... I considered it too but broadly decided against it. The goal of this project as I see it (and I'm absolutely open to counterarguments) is to provide a library that:
* is useful for porting existing libtcod projects to Rust
* easy to use to for people already familiar with libtcod
* be accessible to people not familiar with libtcod, Rust and gamedev

I want to have a safe and idiomatic Rust API as well, but that's a bit less important so long as we keep the common cases safe and easy. I think we're already better than e.g. the Python bindings (their use looks nothing like idiomatic Python code and it's quite easy to trigger segfaults doing common tasks).

As an aside: I dream of a fully Rustic roguelike toolkit that exists as a collection of crates usable individually, but with a good getting started story and distribution (so e.g. a crate that pulls the other ones as dependencies and connects them together, without relying on external libraries for compilation or distribution). With a solid renderer akin to the [BearLib Terminal](http://foo.wyrd.name/en:bearlibterminal) (I find libtcod's drawing complex but limited), path finding, noise functions, FOV, dungeon generators etc. But I don't think this repo is the place for that.

tomassedovic:
Oh a tiny suggestion (but not going to block merging on it): what do you think of renaming `Bsp::new_with_size` to just `Bsp::new`? There's no other constructor.

Libtcod's C++ and C# bindings essentially do that (with constructor), C and Python use the `new_with_size` suffix but I find it a bit superfluous.

Gustorn:
You can still assign raw pointers, you just can't dereference them outside of an `unsafe` block. I feel like there should be a way to produce either a double free or a memory leak.

tomassedovic:
Damn, I forgot about that. So yeah, this is unsafe:

```rust
fn unsafe_pointer() {
    let mut bsp1 = Bsp::new_with_size(0, 0, 100,100);
    bsp1.split_recursive(None, 2, 5, 5, 1.5, 1.5);
    let mut bsp2 = Bsp::new_with_size(0, 0, 100,100);
    bsp2.split_recursive(None, 2, 5, 5, 1.5, 1.5);
        
    bsp2.left().unwrap().tree = bsp1.left().unwrap().tree;
}
```
results in:

    thread 'bsp::test::unsafe_pointer' has overflowed its stack

But what about making the `tree` struct member in *tcod_sys* private and create an unsafe accessor:

```rust
pub type TCOD_tree_t = Struct__TCOD_tree_t;
#[repr(C)]
#[derive(Copy, Clone)]
pub struct Struct_Unnamed29 {
    tree: TCOD_tree_t,
    pub x: ::libc::c_int,
    pub y: ::libc::c_int,
    pub w: ::libc::c_int,
    pub h: ::libc::c_int,
    pub position: ::libc::c_int,
    pub level: uint8,
    pub horizontal: _bool,
}
impl Struct_Unnamed29 {
    pub unsafe fn tree(&mut self) -> &mut TCOD_tree_t {
        &mut self.tree
    }
}
```

That will require an unsafe block to get to that pointer while keeping the parts we care about safe. I've tried that and it seems to work fine.

Gustorn:
Goddamit, that's such a nice and obvious solution. For some reason I always forget about chaning the `bindgen` generated stuff.

tomob:
Click-type-type-type-click-type-type.

Updated.

tomassedovic:
Thanks! And sorry again it took so long.

tomob:
Don't be. I prefer have it made good and solid rather then rushed and sloppy.

PR #201, opened 2015-09-28
Replace external `Time` crate with `std::time`

tomassedovic:
`std::time::Duration` is stable as of Rust 1.3. Using it simplifies our
dependencies and should make interoperability with other code easier.

There is a couple of downsides:

First, the API of the Duration in std is smaller than the one in the
time crate so we need to e.g. do our own conversion to total seconds and
milliseconds elapsed. I'm hoping this will get better in the future or
will be something the time crate will adopt while using
std::time::Duration for the underlying data.

Second, this is a breaking change because it no longer works with the
older (<=1.2) compilers.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,6 @@ path = "src/lib.rs"
 [dependencies]
 bitflags = "0.1"
 libc = "0.1"
-time = "0.1"
 lazy_static = "0.1"
 rustc-serialize = { optional = true, version = "0.3" }
 serde = { optional = true, version = "0.6" }
--- a/examples/samples.rs
+++ b/examples/samples.rs
@@ -21,6 +21,7 @@ use rand::ThreadRng;
 use std::char::from_u32;
 use std::fs::read_dir;
 use std::cmp::{min, max};
+use std::time::Duration;
 
 const SAMPLE_SCREEN_WIDTH : i32 = 46;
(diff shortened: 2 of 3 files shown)


tomassedovic:
Would appreciate peoples' thoughts on this. I think using the data structure defined in `std` is better than an external one, but the downsides (smaller API & support for Rust 1.3 and up only) are real.

Also, if we do go for it, do we lump it with the rest of the changes for our next release or release what we have and keep this for the next one? (I prefer the former, one of the last breaking changes, hopefully)

tomob:
This begs a somewhat philosophical question: how many old versions should the library support; not just for this change, but in general? What's your take on that?

As per the release question: I would include all changes in the release. Rust is still evolving fast (6-week release cycle for a compiler is *crazy* fast IMO) and I see no reason why tcod-rs should evolve with it.

Gustorn:
I think we should support the latest stable Rust build, so I'm in favor of merging this PR. It's a reasonable expectation that users to upgrade within a major version, since backcompat is preserved. This goes for things that are stabilized in the future as well.

Also, I'm in favor of including everything in the new release.

tomassedovic:
I'm in favour of supporting the Stable channel rather than a particular release, but depends on how you look at it, bumping the required rustc version will break people. Not the existing code, necessarily, but what used to compile no longer does.

Once things stabilise a bit, we should probably look at the rest of Rust's ecosystem and follow that.

PR #211, opened 2015-10-27
Fix for crater builds

### LINKED ISSUE #210: Fix the failures on Crater
Crater is a tool that compiles everything on http://crates.io with two version of the Rust compiler to detect breakage.

Tcod is failing, because the build script in tcod-sys is expecting SDL stuff that's not present in the build environment:

https://tools.taskcluster.net/task-inspector/#oHQxc8A-REGyGXfzk37EPQ/
https://tools.taskcluster.net/task-inspector/#KXt0JaJ9QXWWBJsMyttAFg/

See the last few comments on #205 for some discussion around this. Since we only care about the Rust code for Crater runs, we can use its environment variables to detect that and not do anything in build.rs.

Gustorn:
Here's a basic workaround for the crater builds. A few things:

- Should I check for every enviromental variable, or do you think think `CRATER_TASK_TYPE` is enough? 
- This has the annoying property that the user has to manually do a `cargo clean` betwen a crater and a non-crater build. I don't think this is too much of a downside, but I have no idea how to fix it

Closes #210

### DIFF
--- a/tcod_sys/build.rs
+++ b/tcod_sys/build.rs
@@ -3,6 +3,12 @@ use std::path::Path;
 use std::process::Command;
 
 fn main() {
+    let is_crater = option_env!("CRATER_TASK_TYPE");
+
+    if is_crater.is_some() {
+        return;
+    }
+
     let src_dir = env::var("CARGO_MANIFEST_DIR").unwrap();
     let dst_dir = env::var("OUT_DIR").unwrap();
 


tomassedovic:
`CRATER_TASK_TYPE` is enough I think.

Good point on the cargo clean bit. It shouldn't be a problem since people aren't going to be running this under a crater-like environment when doing development.

I'll upload a point release shortly.

tomassedovic:
tcod_sys 3.0.2 is on crates.io. I'll keep an eye on the new crater runs.

PR #213, opened 2016-02-18
Add Windows CI

tomassedovic:
It uses appveyor.com and should provide msvc and mingw coverage on 32
and 64 bit machines.


tomassedovic:
We'll need to fix the windows tests but that'll be much easier with the CI on. Merging.

Issue #222, opened 2016-04-20
Compilation Errors for `tcod-rs`-based Project on `x86_64-pc-windows-gnu`


indiv0:
I'm currently trying to get a project of mine (which uses `tcod-rs`) to compile on Appveyor with `x64_64-pc-windows-gnu`.

It works with MSVC.

It compiles fine on GNU as well, but when trying to run doc tests it gives the following error:

```
    error: linking with `gcc` failed: exit code: 1
    note: "gcc" "-Wl,--enable-long-section-names" "-fno-use-linker-plugin"
    "-Wl,--nxcompat" "-nostdlib" "-m64" "C:\\Program Files
    (x86)\\Rust\\lib\\rustlib\\x86_64-pc-windows-gnu\\lib\\crt2.o" "C:\\Program
    Files (x86)\\Rust\\lib\\rustlib\\x86_64-pc-windows-gnu\\lib\\rsbegin.o" "-L"
    "C:\\Program Files (x86)\\Rust\\lib\\rustlib\\x86_64-pc-windows-gnu\\lib"
    "C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\rustdoctest.IFn9rfSyYfPq\\rust_out.0.o"
    "-o"
    "C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\rustdoctest.IFn9rfSyYfPq\\rust_out.exe"
    "-Wl,--gc-sections" "-nodefaultlibs" "-L"
    "C:\\projects\\tcod-window\\target\\debug\\deps" "-L"
    "C:\\projects\\tcod-window\\target\\debug" "-L"
    "C:\\projects\\tcod-window\\target\\debug\\build\\tcod-sys-621017da41165f97\\out"
    "-L" "C:\\Program Files (x86)\\Rust\\lib\\rustlib\\x86_64-pc-windows-gnu\\lib"
    "-Wl,-Bstatic" "-Wl,-Bdynamic"
    "C:\\projects\\tcod-window\\target\\debug\\libtcod_window.rlib"
    "C:\\projects\\tcod-window\\target\\debug\\deps\\libtcod-a271f719c3238129.rlib"
    "C:\\projects\\tcod-window\\target\\debug\\deps\\liblazy_static-5e6d9f365bf63baa.rlib"
    "C:\\projects\\tcod-window\\target\\debug\\deps\\libtcod_sys-67dc38e7808c55ef.rlib"
[...]

PR #225, opened 2016-06-20
Remove the '.dll' suffix from windows-gnu target

tomassedovic:
This should fix the issue #222 according to the following build run:

https://ci.appveyor.com/project/tomassedovic/tcod-window/build/1.0.5

Looks like MinGW expects to get just the dynamic library's name without
the suffix.


tomassedovic:
The build's failing on the OS X nightly, but it's not been caused by this patch as evidenced by a readme change resulting in the same failure:

https://github.com/tomassedovic/tcod-rs/pull/226
 
So I'm merging this.

PR #228, opened 2016-12-12
Add Console::get_default_background

tomassedovic:
Useful counterpart to the `set_default_background` method.


tomassedovic:
The osx failures are on master, too. See issue #230. I'll merge this as it doesn't seem to introduce new failures (and works fine on Linux and Windows)

PR #229, opened 2016-12-12
Fix the with and height asserts in Console::rect

tomassedovic:
The upstream docs are wrong on this one: `x + width` can be equal to the
console's width as well. It would be impossible to have a rect that ends
at the right edge of the console. And the same goes for height.

In addition, I've split the conditions each into its own assert. When it
fires, it's useful to know which condition in particular failed (was it
the `x` or the `y` check?). This generates much more useful error
messages.


tomassedovic:
The linux and windows jobs passed and the osx ones will fail (see #230), so I'm merging this.

PR #236, opened 2017-03-04
Test msvc breakage

tomassedovic:
this was the first time we saw msvc breaking:

https://ci.appveyor.com/project/tomassedovic/tcod-rs/build/1.0.196

As far as I can tell, this should not have caused any trouble, but let's see if it did anyway.


tomassedovic:
Alright, that wasn't it.

Issue #244, opened 2017-06-28
Map ASCII codes to font?


obsoke:
Hey there. I've been attempting to use graphical tiles in my tcod-rs based game. I found [this wiki article](http://www.roguebasin.com/index.php?title=Complete_Roguelike_Tutorial,_using_Python%2Blibtcod,_extras#Using_Graphical_Tiles) outlining how this can be done with the Python version of tcod but I've gotten stuck replicating these instructions with these Rust bindings.

I'm loading the custom font with the following form of the `Root`'s builder:

```rust
let mut root = Root::initializer()
    .font("TiledFont.png", FontLayout::Tcod)
    .font_type(FontType::Greyscale)
    .font_dimensions(32, 10)
    .size(SCREEN_WIDTH, SCREEN_HEIGHT)
    .title("Roguelike")
    .init();
```

But when it comes to the wiki's next step of using `map_ascii_codes_to_font`, I get stuck. I searched the repository but it seems like the only reference to this method is in the `tcod-sys` crate which does not seem to be publicly exposed. Additionally, the method doesn't seem to be called by any publicly available functions.

Am I missing something or is there no way to accomplish this at this moment? If it's currently missing, do you have any thoughts on how this can/should be implemented? Perhaps keeping it within the `Root`'s builder pattern as with custom fonts? I'm happy to help out with a PR!

Gustorn:
I looked into the `map_*_to_font` functions and it seems like they should be called after the `Root` console has been initialized. This would mean that they need to be added as methods taking `&mut self` to `Root` [here](https://github.com/tomassedovic/tcod-rs/blob/master/src/console.rs#L198). 

Unfortunately I don't have the time to implement them right now but you can look at the existing functions at the linked location if you'd like to send a PR.

tomassedovic:
Yep. And to hopefully clarify your confusion: `tcod_sys` contains the direct unsafe bindings to libtcod. They were generated automatically and should have all the libtcod functions with identical names.

The `tcod` crate uses these to provide higher-level bindings that are hopefully safer and look more like Rust and less like C. These are not complete yet and some things such as the `map_ascii_codes_to_font` function are not there yet.

If you add "tcod-sys" in your `dependencies` and use `extern crate tcod_sys`, you should be able to call the raw functions within an `unsafe` block.

So the line in the tutorial would look something like this:

```rust
unsafe {
    tcod_sys::TCOD_console_map_ascii_codes_to_font(a, 32, 0, y);
[...]

PR #245, opened 2017-06-30
Implements the "map_*_to_font" series of methods to Console::Root

obsoke:
As discussed in #244, Adds the following methods to `Root`:

* `map_ascii_code_to_font`
* `map_ascii_codes_to_font`
* `map_string_to_font`

No tests included in this PR (yet) as I'm unsure as to how to go about testing these methods . It seems like you need to use your eyes to see whether the correct font character was rendered.

Please let me know if anything is missing / needs fixing!


src/console.rs
```diff
@@ -345,6 +345,52 @@ impl Root {
...
+                nb_codes,
+                font_char_x,
+                font_char_y
+            );
+        }
+    }
+
+    /// Maps ASCII codes from a string to consecutive characters in a bitmap font.
+    pub fn map_string_to_font(&self,
+                              s: &'static str,
```
Gustorn: You don't need the `'static` lifetime here (and it actually limits some valid use-cases).

Gustorn:
All of these should take `&mut self` instead of `&self`: if you look at the [original code](https://github.com/tomassedovic/tcod-rs/blob/master/tcod_sys/libtcod/src/sys_sdl_c.c#L215) there's some mutation going on which we should reflect in our API.

tomassedovic:
Thanks! I can merge this as is, but there are two ways of improving it.

First, explain the function parameters in the documentation (the `///` comments turn into the docs). It's mostly self-explanatory here, but still.

Second, you're right there's no easy way to test this automatically, but we could write an example. That's good because it shows how to use the feature and makes sure we don't break the API by accident.

You can see this:

https://github.com/tomassedovic/tcod-rs/blob/master/examples/chars.rs

And run it as:

    cargo run --example chars

obsoke:
Ah, an example is a good idea in lieu of a decent testing method. I will add that & expand the doc comments. Thanks to both of you for the feedback so far!

obsoke:
I've added an explanation of function arguments, as well as a working example. I have to admit that the example is a bit contrived. Personally, I think showing off the use of a graphical file font (my original use case) would be neat but this would require adding a new font to the project. What do you folks think? Is it worth having a cooler example, or is this Good Enough? Also, please let me know if anything needs fixing up.

tomassedovic:
Thanks!

I agree that adding a custom font would be much better (and is what these functions are designed for) and so long as there are no licensing issues with the font, I'd prefer that. Something like creative commons or public domain should be fine.

But if you want to merge this now (it looks good to me), we can do that and add a better example in a later pull request. Whichever you prefer.

obsoke:
angband16x16.bmp is in public domain, [according to its creator](http://angband.oook.cz/forum/showpost.php?p=316&postcount=16).

christopherdumas:
In my issue #247, I have a tileset that looks pretty nice (in my opinion). It was drawn by a friend of mine. You could use that, although it is possibly broken or incorrectly formatted (see the issue).

christopherdumas:
I would really like this merged soon, because right now I'm writing a game that is using the `map_*_to_font` functions in unsafe mode and it would be nice to not have to use unsafe.

obsoke:
Is there anything else you folks need in this PR? I've updated it to have a less contrived example & added a public domain graphical tileset.

tomassedovic:
I'm really sorry it took me so long. This looks great, merging it now.

PR #256, opened 2017-11-16
libtcod 1.6 support

### LINKED ISSUE #249: Support for libtcod 1.6.3+
Any plans to support v1.6+?

Gustorn:
Please see Gustorn/tcod-rs#1 for a pretty annoying warning message that I don't know how to get rid of. Also this is - as of yet - untested on Windows and OS X. 

- [x] The library builds on Travis under Linux
- [x] The library builds on Travis under OS X
- [x] The library builds on AppVeyor under Windows with the MSVC toolchain
- [x] The library builds on AppVeyor under Windows with the GNU toolchain
- [x] Fix undefined symbols warnings when running a built application

Closes #249


Gustorn:
I think something's wrong with Travis. @tomassedovic have you seen this error before? It fails before ever getting to the build. The AppVeyor failure is totally fair I'll try to look into it on Windows.

Tyruiop:
I'm sorry I can't directly help with the windows part, but I have been looking about using the autotools (which I think would also simplify the build for windows, no? As well as the issue regarding opengl/sdl2) but I am currently having issues with the libz (the symbols are missing in the libtcod I output).

Gustorn:
I think using the `cc` crate is our best best, I'd rather not run a different application in the cargo build. I'll add a checklist in the original comment so we can track what needs to be fixed before this gets merged.

Tyruiop:
I agree with the cc crate, just for note though: I got it working with
```
        build_libz(libz_sources);

        Command::new("autoreconf")
            .arg("-i")
            .current_dir(&Path::new("libtcod/build/autotools/"))
            .status().unwrap();

        Command::new("sh")
            .arg("./configure")
            .current_dir(&Path::new("libtcod/build/autotools/"))
            .status().unwrap();

        Command::new("make")
            .current_dir(&Path::new("libtcod/build/autotools/"))
            .status().unwrap();

        Command::new("cp")
            .args(&["libtcod/build/autotools/.libs/libtcod.so.0.0.0", &format!("{}/libtcod.so.0", dst.display())])
            .current_dir(&Path::new("./"))
            .status().unwrap();

        pkg_config::find_library("sdl2").unwrap();
        println!("cargo:rustc-link-lib=static=z");
```

For the linux build and it got rid of the XF86 symbols debug messages.

That leads to the following question for me: how about not shipping libtcod with the bindings and letting the user build it? Of course that means that the bindings need to be kept up to date regularly, but I think many distributions have a libtcod package, windows has binaries and mac has.. ok I don't know what mac has.

Gustorn:
We are letting the user build it, only it's automated. I think it's still a better idea to include it, especially on Windows. I'll look at what autotools does and see if I can reproduce it using `cc`.

Gustorn:
I fixed the Linux and OS X builds but I'm pretty stumped with the Windows one. I'll get back to it later today.

Gustorn:
Only Windows GNU and the weird warnings are left. If anyone has any idea how to fix them feel free to send a PR agains my branch, I'll merge it and it'll show up here.

Tyruiop:
If you mean the `DEBUG: Failed loading XShmCreatePixmap: /usr/lib/libX11.so.6: undefined symbol: _XShmCreatePixmap` messages, I will investigate how to reproduce the autoconf building instructions, since they cleared them

Gustorn:
@tomassedovic just out of curiosity: is there a reason why we don't link the libtcod statically? That's what all the other `*-sys` crates seem to be doing and it would clean up the build file quite a bit

Gustorn:
I tried building/linking it statically, no changes with the undefined symbol warnings.

tomassedovic:
@Gustorn I kind of thought static linking wouldn't make a lot of sense if we also didn't build SDL statically, and that was apparently discouraged.

If it fixes things, I'm all for it.

@Tyruiop we did use to use the provided make etc. back in the day, but it make debugging build failures a nightmare so one day I sucked it up, read what exactly the build process was doing and reproduced it in `build.rs`. I'd prefer if we could keep doing that rather then calling out to external tools if possible. It makes things easier for the users of the crate.

Gustorn:
I can still simplify the build using shared libraries, it's just not as nicely supported by `cc` out of the box. The main idea is that the separate `Building  *.o files` is largely unnecessary: since the library basically doesn't change we don't really care about incremental compilation.

This would cut out a pretty large portion of the current build file and nicely simplify the rest. Any thoughts on this change?

tomassedovic:
@Gustorn I'm all for it. As long as this doesn't make things harder for our users (and afaics it shouldn't), I'm good with removing code :-)

zetashift:
Is this still doable? Anything I can do to help with this?

tomassedovic:
@zetashift I think so! @Gustorn would probably be the best to tell you the current status, but it seems that we just need to fix the mingw (pc-windows-gnu) builds. And that he's looked into compiling libtcod statically -- if we can do that, all the better.

Basically, fix the two failing appveyor checks (click on Details to see their logs) and we're good.

It will still be a while before I get to properly test and review it (the change is by its nature massive), but if it all works everywhere, I'm happy to get it in.

Sadly, I can't offer a lot of mentorship on this, but the first steps would be to take the pull request, try to build it on Windows under mingw, try to build upstream libtcod under mingw, figure out what's wrong and fix in our build.rs.

Gustorn:
Unfortunately I don't have the time to investigate the two remaining issues:

1. Undefined symbol warnings on Linux when running the built binary (but it works)
2. Windows hold failure

If someone solved them and makes a PR against my branch I'll pull and test it and then this can get merged as well.

Gustorn:
Thanks to @L3nn0x for getting the Windows GNU build working. The last remaining issue is the undefined symbol warnings under Linux.

L3nn0x:
I'll try to take a look at that last issue and static compilation as well, see if I can come up with something.

Gustorn:
I actually banged my head against that wall for quite a bit and couldn't come up with a working solution. I converted the whole `build.rs` to use static linking to no avail. In hindsight I probably should've documented what I tried but too late for that unfortunately. As mentioned earlier in the comments, I'd recommend starting with the autotools build to see where it differs from ours.

lucanLepus:
@Gustorn I've just looked into this and it appears to me those "warnings" are mere debug messages which are suppressed in the autotool build by the `-DNDEBUG` define.

adding line 125 in build.rs
`config.define("NDEBUG", None);`
gets rid of them.

L3nn0x:
Thanks @Gustorn ! Just added the commit and it does remove the messages.

Gustorn:
Thanks for the investigation/solution, I merged the PR into my branch, if it passes on Travis then I'm OK with merging this.

Gustorn:
I think this one's ready but I don't want to merge a breaking change without @tomassedovic's permission (even though the major version bump correctly reflects the status quo).

lucanLepus:
i have some more thoughts about the build,

first off, while i know my way around git my github skills are lacking to the point where i don't know how i would propose changes to a standing pull request, so feel free to teach me :-)

1. the `NDEBUG` flag is right now only passed when building on linux, should probably also be done for other targets.
2. there are still compiler flags to link against OpenGL, which should be superflous when using `NO_OPENGL`. While probably doing no harm i can imagine unnecessary build-failures when OpenGL is not found

i've pushed those changes to my fork [here](https://github.com/lucanLepus/tcod-rs/commits/tcod-16). It builds fine on linux but i cannot test anywhere else.

L3nn0x:
I don't think we have the same problem with the other platforms, that's why I didn't put the `NDEBUG` flag elsewhere. Good catch for the `NO_OPENGL` flag. I suggest you open a PR against https://github.com/Gustorn/tcod-rs/tree/tcod-16-upstream-rebase. @Gustorn can have a look at it then.

lucanLepus:
i guess it comes down to what kind of build of libtcod should be provided here.
without `NDEBUG` it is basically a debug build, with assertions and verbose logging enabled.

if, which i assume, we want to provide a release build here, `NDEBUG` should be defined in every build.

mipli:
Been using this branch for a while now on my own game, and it's been working fine. Only issue I had when changing was the new key press usage in SDL2, which they talk about here https://libtcod.readthedocs.io/en/stable/_source/upgrading.html  But the only change I had to make for that was to add a test for `code: KeyCode::Text` all the places I also checked for `printable...`. Easy once I knew what the issue was.

That means that the tutorial will need to be updated as well once this is merged. 

The Roguelikedev community over at Reddit is starting a [follow-along roguelike tutorial ](https://www.reddit.com/r/roguelikedev/comments/8ql895/roguelikedev_does_the_complete_roguelike_tutorial/) on the 19th of June, and at least some people are going to be using Rust and probably this library and accompanying tutorial. So, if this is merged now the tutorial should get updated before then as well so new users don't get confused.

L3nn0x:
That sounds nice! I'll probably try to participate if I have the time.

L3nn0x:
@Gustorn @tomassedovic I've been testing the branch on linux and it works. Any chance one of you can take a look and merge/comment?

Thanks!

tomassedovic:
My deepest apologies for taking so long! Thank you all for doing this! I'm merging it now.

Here's a few things I'd like to see happen before we release a new version on Crates.io though:

* Readme needs updating:
  - mention we're using 1.6 now
  - I've pushed a `libtcod-1.5.1` branch for folks who want to stay on it -- we should document that
  - @Tyruiop, @L3nn0x and @lucanLepus need to be added to Cargo.toml as contributors
  - we must mention that SDL2 is now a dependency
* Cargo.toml version needs updating
* possibly other stuff I'm missing? But that's the most important bits

Also, @Tyruiop, @L3nn0x and @lucanLepus since you've all contributed to this repo, I can give you merge rights if you want. That helps this little project going even when the original authors have moved on or don't have that much time these days. Please let me know if that's something you'd be interested in.

tomassedovic:
Ooh, also! This is a pull request number 256! How fitting!

tomassedovic:
Thanks @L3nn0x, you should have received an invite!

What you said is perfectly okay! I don't have a ton of time either (as evidenced by the time it took me to merge this). But hopefully the more people the greater chance this project won't die out of neglect :-).

lucanLepus:
@tomassedovic yes, i would not mind having merge rights here, but like all of you i don't have a lot of time on my hands, so i probably would not use the rights much.

tomassedovic:
Thanks @lucanLepus, you should receive an invitation.

Totally understand the time issue, we're all in the same boat.

Issue #262, opened 2018-08-28
option to build libtcod statically?


laanwj:
It's currently a bit of a pain to run examples and applications using the library at the moment due to the `libtcod.so` dependency at run time.

I think statically linking the internal libtcod (e.g. as a `.a`) could avoid this and make the resulting executables self-contained, of course apart from system libraries such as SDL.

tomassedovic:
I agree, this would be a great usability win.

Unfortunately, I don't really have the time and knowledge to investigate this at the moment. Do you? I'd be happy to help in any way I can and I would love to accept a pull request that implements it.

laanwj:
Sure!

I can't promise how soon, but I'm willing to look at this.

laanwj:
I looked into it, it's easy enough to make it link to an `.a` instead of an `.so`, as well as to add `println!("cargo:rustc-link-lib=dylib=SDL2");` to specify SDL2 as a dependency for linking libraries.

However, it looks like rust insists on passing the order of the linker arguments as `"-l" "SDL2" "-l" "tcod"` which is the wrong way around, so I get screens full of symbol errors. Nothing that I tried could make it work.

tomassedovic:
The PR above is merged, I'm closing this issue. Thanks so much for doing this @laanwj!

PR #264, opened 2018-08-29
Build static library for libtcod on Linux

laanwj:
Build static library for libtcod on Linux

Tell rust to link the static version of libtcod on Linux.
This avoids issues about where to put `libtcod.so` at run time by including the code in the executable itself.

This is my attempt at #262. I think the same should work for other operating systems but I am unable to test this right now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,17 @@ on your distro.
 
 You can also check the [official libtcod build instructions for Linux](http://roguecentral.org/doryen/data/libtcod/doc/1.5.2/html2/compile_libtcod_linux.html?c=true).
 
+#### Building a dynamic library
+
+By default, `tcod-rs` will build the library statically on Linux as including
+the code into the executable is usually more convenient. To build a dynamic
+library specify the `dynlib` feature for `tcod_sys` in `Cargo.toml`
+
+```
+[dependencies.tcod_sys]
+version = "..."
+features = ["dynlib"]
+```
 
 ### Building on Windows (with MSVC)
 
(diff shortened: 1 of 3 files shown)


laanwj:
sure!

what is the way to do build options in rust, a crate feature?

L3nn0x:
I'm not entirely familiar with the build system in rust, but the [build profile](https://doc.rust-lang.org/book/second-edition/ch14-01-release-profiles.html) could be interesting to look into.

L3nn0x:
I've been looking into this, and we could use a feature for it.
Example:
```
if cfg!(target_os = "macos") || cfg!(target_os = "ios") {
    println!("Think Different!");
}
```
Link: https://doc.rust-lang.org/book/first-edition/conditional-compilation.html

laanwj:
I also think a feature is probably the way to go. 

I suppose static compilation makes sense as default, as it is most easy to use, then dynamic as feature?

Build profiles seem to be for the entire application ("release", "debug" etc), I don't think we want to mess with requiring that at a crate level.

L3nn0x:
Yes I agree, features seem better suited for our use case.
I'm all for statically linked by default. I'll invoke @tomassedovic for that, but go for it until he replies.

L3nn0x:
@laanwj any updates on this?

laanwj:
yes, still intend to work on this soon!

laanwj:
OK — updated to use static linking by default unless the feature `dynlib` is passed to `tcod_sys`, then it will do the same thing as before. Mentioned this in the README.md.

L3nn0x:
Thanks!

Issue #259, opened 2018-07-02
Mark part of the library as Send


L3nn0x:
When trying to use the library in a multi-threaded environment, most of the structs aren't Send because they contain a raw pointer. Would it be possible to marke them as Send? I've been doing it in my local branch and it works, but I'm not sure if it's the correct way of doing it.

tomassedovic:
Hm. So the problem is that libtcod (the underlying C library) itself is not thread-safe. Things may work, but they may also develop hard-to-debug problems randomly.

I *think* most of the structs could just be sent across threads, but only if the raw pointers are unique (multi-threading is not my strongest suite). And I'm not sure our library actually guarantees that. So e.g. if you created two structs that point to the same underlying libtcod Map, it would not be okay. This is why Rust doesn't implement Send on raw pointers.

If you want to send stuff across threads, you should be able to wrap the struct in a [std::sync::Mutex](https://doc.rust-lang.org/stable/std/sync/struct.Mutex.html).

This is something we could add to the library, but it would require a significant amount of effort.

L3nn0x:
Wrapping the struct in both an Arc and a Mutex only guarantees Send if the underlying struct is Send. Unfortunately raw pointers aren't. I know that libtcod isn't thread safe, but maybe we could make the structs movable between threads? It wouldn't be adding thread safety, just being able to move the struct between two threads.

L3nn0x:
I completely agree with you, it's not the best solution.
I've been looking a little into it and I think the best solution would be to store the raw pointer in a box. I'm not sure how easy it would be, but there is a `Box::from_raw()` and a `Box::into_raw()` methods that look like they could do the job.

tomassedovic:
Oh that sounds like a great idea! I should have thought of that.

Reading the `from_raw` docs though, it says that the only valid pointer to pass there is another one created by a Box::into_raw though :-(.

https://doc.rust-lang.org/std/boxed/struct.Box.html#method.from_raw

I'm not sure what that means in practice, but it might mean that on certain platforms or std implementations things would not work properly, sigh. Maybe our best bet is to just implement Send on the structs after all.

L3nn0x:
I'm pretty sure this is because Box frees the memory pointed to when dropped. That's why I was looking into `Box::into_raw()` as well, to remove the pointer from the Box before it is dropped. The problem I see is that it's not possible to call this method from inside the `Drop trait` simply because it consumes the box.
[...]

PR #265, opened 2018-09-17
Sync/Send for Map

fadeddata:
I needed this change so that I could add `Map` as a `Write<'a Map>` resource using specs (https://github.com/slide-rs/specs). It's been working great and I have not witnessed any sort of issues. Is there any reason this couldn't be merged?

Thanks!

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -1,3 +1,4 @@
+use std::sync::{Arc, Mutex};
 use bindings::ffi::{self, TCOD_fov_algorithm_t};
 use bindings::{AsNative, c_bool};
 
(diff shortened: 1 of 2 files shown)


fadeddata:
@L3nn0x I hope I did that right. Let me know and I'll make any required changes.

Gustorn:
I wouldn't add `Arc<Mutex<_>>` in the library itself. You can make it send and if people need to access it from multiple threads then they can just use an `Arc<Mutex<Map>>`

fadeddata:
What about starting a `safe` module that wraps all of the API's we might want to have wrapped in `Arc<Mutex<_>>` ?

L3nn0x:
I don't know where @tomassedovic wants to go with this library, but it could be a possibility. Feel free to open an issue about it so we can discuss it more in depth.

Gustorn:
I feel like a big part of the Rust philosophy is zero-cost abstractions, adding `Arc<Mutex<_>>` around everything is way too restrictive:

1. For one, users of the library might not want to use the types in a multithreaded context
2. They might use multiple threads but opt for a different thread synchronization method

In both of these cases, users of the library would be paying for the thread-safe abstractions for no reason. 
I think wrapping your own types if you need to is a perfectly valid way to go about things: consider that the Rust standard library isn't thread-safe by default either (with some notable exceptions of course).

L3nn0x:
@fadeddata just realized I already had an issue about it open here: https://github.com/tomassedovic/tcod-rs/issues/259
I was looking into using Box to store the raw pointers, but I haven't had the time to do a poc. That would make the library more "rustacean".

PR #269, opened 2018-11-14
README.md: mention that version 1.6.3 is wrapped

laanwj:
This was updated in #256, since then https://github.com/tomassedovic/tcod-rs/blob/master/tcod_sys/libtcod/README.md mentions 1.6.3, however the README.md incorrectly still mentioned 1.5.2.
(the project description on github also does)

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ useful functionality such as:
 
 
 This project provides [Rust](http://www.rust-lang.org/) bindings for libtcod
-v1.5.2.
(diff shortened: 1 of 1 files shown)


laanwj:
Might also want to update the documentation URL to https://libtcod.readthedocs.io/en/stable/, though I think that's for the latest stable, there is no specific one for 1.6.3 as far as I know.

L3nn0x:
Thanks!

Issue #267, opened 2018-10-15
Black screen on OSX Mojave


kikito:
Hi,

I'm using (learning, really) tcod-rs on OSX.

Everything was working fine until I updated to Mojave. After that, all I see is an empty black screen.

I have tried `tcod-rs/examples/text.rs` too in order to discard problems in my code and I confirm that I also see a black screen when I run it.

I have installed libtcod using `brew install libtcod`. Here's what `brew` says about it:

```
$ brew info libtcod
libtcod: stable 1.6.4 (bottled)
API for roguelike developers
http://roguecentral.org/doryen/libtcod/
Conflicts with:
  libzip (because both install `zip.h` header)
/usr/local/Cellar/libtcod/1.6.4 (141 files, 2MB) *
  Poured from bottle on 2018-10-15 at 11:29:03
From: https://github.com/Homebrew/homebrew-core/blob/master/Formula/libtcod.rb
==> Dependencies
Build: autoconf ✔, automake ✘, libtool ✔, pkg-config ✔
Required: sdl2 ✔
==> Analytics
install: 19 (30d), 59 (90d), 330 (365d)
install_on_request: 19 (30d), 51 (90d), 275 (365d)
build_error: 0 (30d)
```

kikito:
Googled a little bit more. It seems this might be a problem with libsdl.

* Similar bug: https://discourse.libsdl.org/t/macos-10-14-mojave-issues/25060/6 (my screen did *not* draw anything after moving the window around with the mouse)
* Bugzilla: https://bugzilla.libsdl.org/show_bug.cgi?id=4272
* Patch for fixing the previous sdl bug (I don't know how related that is with the one I am experiencing) https://hg.libsdl.org/SDL/rev/55489adbb75c

I will try to recompile libsdl with that patch to see if that helps. But it will take me a while.

L3nn0x:
We only provide bundled sdl2 libs for windows, so your best bet would be to either update your package or recompile the library with the fix.

ttracey93:
@kikito Did you get this working?

kikito:
@ttracey93 no dice. Icculus has this thread on twitter listing things that Mojave broke on OpenGL (and thus SDL):

https://twitter.com/icculus/status/1054634481812992000

I would expect these issues to be fixed  eventually in SDL. I digged in the SDL issues page and found this one:

https://bugzilla.libsdl.org/show_bug.cgi?id=4329

[...]

PR #270, opened 2018-12-04
Fix SDL2 rendering in Mojave

rjungemann:
Related to https://github.com/tomassedovic/tcod-rs/issues/267

Basically, SDL2 on OS X requires you to `SDL_PumpEvents()` before setting the window size, otherwise you get a blank screen until it receives an event, such as a keyboard event or window resize event.

See also: https://discourse.libsdl.org/t/macos-10-14-mojave-issues/25060

### DIFF
--- a/tcod_sys/libtcod/src/sys_sdl2_c.c
+++ b/tcod_sys/libtcod/src/sys_sdl2_c.c
@@ -289,6 +289,8 @@ static void create_window(int w, int h, bool fullscreen) {
 #endif		
 		if (TCOD_ctx.renderer == TCOD_RENDERER_SDL ) {
 			window = SDL_CreateWindow(TCOD_ctx.window_title,SDL_WINDOWPOS_CENTERED,SDL_WINDOWPOS_CENTERED,w*TCOD_ctx.font_width,h*TCOD_ctx.font_height,winflags);
+			SDL_PumpEvents();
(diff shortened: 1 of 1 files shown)


L3nn0x:
Thanks!
I'm wondering if it wouldn't be better to make the PR on libtcod directly: https://bitbucket.org/libtcod/libtcod/overview

Tell me if you want to do this and I'll update the repo with the latest libtcod.

rjungemann:
Oof I've never delved into BitBucket/Mercurial but I'll give it a shot.

rjungemann:
@L3nn0x this will get fixed once SDL2 2.0.10 is out but I'll try and figure out how to make the change over at the tcod repo.

L3nn0x:
Will it fix the pump event problem as well? If yes, you don't need to update tcod I'll just update the sdl version we link against.

tomassedovic:
I'm happy to merge this to help unblock people experiencing these issues.

@rjungemann are you okay with that? It's been several months so I'm basically just checking in to see if you're still fine with this being the right fix until SDL2 2.0.10 is out in homebrew?

tomassedovic:
As of today (9th June, 2019) the SDL 2 in Homebrew is still 2.0.9:

https://formulae.brew.sh/formula/sdl2

I'm merging this and I'll publish a new tcod-rs version with the accumulated fixes.

PR #275, opened 2019-01-17
Update version of the SDL dev bindings required

muhuk:
v0.13 seems to require SDL2 instead of SDL1.2.


tomassedovic:
Good catch, thanks!

PR #283, opened 2019-05-04
make Color::new a const function

ElectricCoffee:
Makes it so that it's possible to write

```rust
const LINK_COLOR: Color = Color::new(0x00, 0xCC, 0x00);
```

Saves the user from using the struct directly.


tomassedovic:
Thanks!

Issue #284, opened 2019-05-06
Make Map impl Clone


nsmryan:
I've run into some situations where it would be nice to copy a Map struct, and there is no way to do this currently without calling ffi functions.

It seems like Map could impl Clone and call TCOD_map_copy to allow new maps a little more ergonomically.

I can make  a pull request for this if you want. The code I came up with is:
```rust
impl Clone for Map {
    fn clone(&self) -> Self {
        let (width, height) = self.size();
        let mut new_map = Map::new(width, height);

        unsafe {
            ffi::TCOD_map_copy(*self.as_native(), *new_map.as_native());
        }

        new_map
    }
}
```

tomassedovic:
That looks good to me! Yes please, if you put up a PR, I'd be happy to review it.

tomassedovic:
Closed via #285

PR #285, opened 2019-05-17
Add 'impl Clone' for Map using libtcod ffi function

nsmryan:
This change adds an implementation of Clone for the Map type using the TCOD_map_copy function from tcod.

This addresses issue #284


tomassedovic:
Thanks!

PR #286, opened 2019-05-19
Added Send to Map

abesto:
In the vein of #274, this is basically a lighter version of #265 until #259 is "properly" resolved. (User code will still need to `Arc<Mutex<Map>>`, which I think is an OK trade-off at this point)

Use-case: using Specs, I want to be a good little boy-scout, and have a dedicated `FovSystem` update a `Map` resource that is then read by the `RenderSystem`.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -84,6 +84,10 @@ impl Drop for Map {
     }
 }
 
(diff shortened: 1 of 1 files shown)


tomassedovic:
Thanks!

PR #302, opened 2019-11-04
Add HeightMap wrapper for TCOD_heightmap_t

alexschrod:


src/heightmap.rs
```diff
@@ -0,0 +1,306 @@
...
+impl HeightMap {
+    pub fn new(width: i32, height: i32) -> Self {
+        Self {
+            height_map: unsafe { ffi::TCOD_heightmap_new(width, height) },
+            width,
+            height,
+        }
+    }
+
+    pub fn values(&self) -> &[f32] {
```
L3nn0x: If we provide a way to get the values, we might want to have a way to get the width and the height as well. In any case I'm not sure if we would have a use case for providing the raw values to the user.

alexschrod: It is essentially a free operation, so I figured I'd throw it in there. If you're working on a large number of contiguous values, I would assume it's faster than having to call `get_value` for each coordinate. But I can remove it if it's undesirable.

L3nn0x: I'm fine with keeping it, but there is no way to access the width nor height, so that slice won't really be usable in the end. Either add a way to get the width/height as well, or remove that method altogether I'd say.

alexschrod: Alright, added width and height getters.

PR #317, opened 2021-01-25
Protect against segfault and clean up warnings

### LINKED ISSUE #297: Fix deprecation warnings on latest Rust

### LINKED ISSUE #310: Segfault running Dijkstra path finding example
```shell
$ cargo run --example dijkstra-path-finding
   Compiling tcod-sys v6.0.0 (/home/shadower/code/tcod-rs/tcod_sys)
   Compiling tcod v0.15.0 (/home/shadower/code/tcod-rs)
    Finished dev [unoptimized + debuginfo] target(s) in 7.46s
     Running `target/debug/examples/dijkstra-path-finding`
Starting from: (4, 6)
Walking to: (4, 6)
Walking to: (3, 5)
Walking to: (2, 4)
Walking to: (1, 3)
Walking to: (0, 2)
Walking to: (1, 1)
Arrived at the destination!

Starting from: (0, 6)
Walking to: (0, 6)
Walking to: (1, 5)
Walking to: (0, 4)
Walking to: (1, 3)
Walking to: (0, 2)
Walking to: (1, 1)
Arrived at the destination!

fish: “cargo run --example dijkstra-pa…” terminated by signal SIGSEGV (Address boundary error)
```

This happens both with the current tcod-rs release (0.15.0) and with: https://github.com/tomassedovic/tcod-rs/pull/309

I suspect that means the error is probably in our bindings rather than libtcod itself, but maybe it is just a long-standing issue in the original codebase.

ojhp:
Fixes #310 - Segfault appears to be in the TCOD library, it can't handle reversing an empty list in the Dijkstra example finding a path from (0, 0) to (0, 0). Added a protective check around the call the the underlying library. The A* implementation works differently enough to be unaffected from what I can see.

Fixes #297 - `...` ranges replaced with `..=` and unneeded brackets removed. Also updated lazy_static to latest version which resolves a deprecation warning generated from a macro in there.

I'll leave this open for a week or so before merging to give chance for comments, none of this seems urgent until the deprecations become breaking.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,7 +35,7 @@ path = "src/lib.rs"
 
 [dependencies]
 bitflags = "0.1"
-lazy_static = "0.1"
+lazy_static = "1.4"
 rustc-serialize = { optional = true, version = "0.3" }
 serde = { optional = true, version = "1.0" }
 serde_derive = { optional = true, version = "1.0" }
--- a/examples/samples.rs
+++ b/examples/samples.rs
@@ -546,7 +546,7 @@ impl Render for NoiseSample {
 
         if let Some((_, Event::Key(key))) = event {
             match key.printable {
-                '1'...'9' =>
+                '1'..='9' =>
                     self.func = {
                         let number = key.printable.to_digit(10).unwrap() as u8;
                         NoiseFunction::from_value(number - 1)
@@ -778,7 +778,7 @@ struct PathSample<'a> {
 }
 
 const TORCH_RADIUS : f32 = 10.0;
-const SQUARED_TORCH_RADIUS : f32 = (TORCH_RADIUS*TORCH_RADIUS);
+const SQUARED_TORCH_RADIUS : f32 = TORCH_RADIUS*TORCH_RADIUS;
 
 static SMAP : [&'static str; 20] = [
     "##############################################",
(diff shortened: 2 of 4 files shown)


tomassedovic:
Great stuff, thank you!

